PR #4, opened 2018-08-30
Add openshift resource parser

aidenkeating:
@priley86 This is a small change, including the `openshift-resource-parser` library which can be used to get a pretty-formatted version of the provisioned services.

The library has two functions:
* `getProvisionedMWService` - Get a single ProvisionedService by name
* `listProvisionedMWServices` - List all ProvisionedServices in a namespace

Docs are available at: https://integr8ly.github.io/openshift-resource-parser/

An example of using the library would be.

```js
import OpenShiftResourceParser from "openshift-resource-parser";

const myAuthToken = "myAuthToken";
const resourceParser = new OpenShiftResourceParser({ openshiftURL: "https://mycluster.com", mockData: false });

resourceParser.listProvisionedMWServices(myAuthToken, "eval")
  .then(provisionedServiceList => resourceParser.getProvisionedMWService(myAuthToken, "eval", provisionedServiceList[0].name))
  .then(provisionedService => {
    console.log(provisionedService);
  })
  .catch(err => console.error(err));
```
And in mock mode.
```js
const mockResourceParser = new OpenShiftResourceParser({ openshiftURL: "https://mycluster.com", mockData: true });
mockResourceParser.listProvisionedMWServices(myAuthToken, "eval")
  .then(provisionedServiceList => mockResourceParser.getProvisionedMWService(myAuthToken, "eval", provisionedServiceList[0].name))
  .then(provisionedService => {
    console.log(provisionedService);
  })
  .catch(err => console.error(err));
```

Mind taking a look?


priley86:
thanks! Will review it shortly...

priley86:
so I don't have any issue w/ this @aidenkeating - seems like a logical path to retrieve middleware service resources. Here is some integration thoughts I am having now:

* it may be nice to hook service responses into the Redux store in some way (but Redux is by no means required). We have started writing services in this project as vanilla es6 methods like the one [here](https://github.com/integr8ly/tutorial-web-app/blob/development/src/services/threadServices.js). It can be called from Redux actions or directly from a page/component. Redux just gives us access to that state globally after the reducer inserts the resulting payload to the Redux store...

* if you need to make use of environment variables for driving auth configurations, whether build time or runtime env variables, we had been exposing them in the `.env` files in the project root. By default, most node apps will respect these [see this article](https://medium.com/@tacomanator/environments-with-create-react-app-7b645312c09d). You can see how we are able to inject a variable w/ the commit hash [here](https://github.com/integr8ly/tutorial-web-app/blob/development/package.json#L56) into the `.env` file at build time, and then the UI can reference these variables automatically. I am writing out all of the `process.env` variables now and you can currently see them when inspecting the UI console [here](http://react-demo-app-test-app.6923.rh-us-east-1.openshiftapps.com/#/) and `REACT_APP_UI_COMMIT_HASH` is parsed into the UI...just some ideas to share about dynamic env variables...

* as far as injecting the service data into the ascii doc/react templates, I think we should just inject them as attributes after the service resources are returned. If they are ascii docs, we can just add them to the `attributes` prop (an object) and pass them in once they become available... https://github.com/integr8ly/tutorial-web-app/blob/development/src/components/asciiDocTemplate/asciiDocTemplate.js#L21

priley86:
@dlabaj any additional concerns? 

@aidenkeating let us know when you'd like this PR merged or you'd like to iterate a bit more first...

dlabaj:
@aidenkeating Was just wondering about the ProvisionedService object.  The UI needs a description of each of the services.  Should we add that property to that object?  Is this something the server will be okay storing for us?

PR #6, opened 2018-08-30
Add a simple server with a dynamic config.js endpoint.

david-martin:
The local development flow remains unchanged with this commit.
The below can still be used.

```
yarn install;yarn start:dev
```

What has changed is what gets started with above command.
A simple node.js express server (served from `server.js`)
starts on port `5001`.
The react dev server proxies any requests it can't deliver statically
to this express server.

Currently, there is only 1 endpoint in this express app `/config.js`.
This endpoint returns some configuration to be used when making API
calls to an OpenShift cluster:

* openshiftAPIUrl (configured from OPENSHIFT_MASTER_URL env var)
* accessToken (as sent on from the oauth proxy after successful OAuth flow)

In a production scenario, the express server will replace where nginx is
currently used (future PR to update openshift templates).
The express server will serve the compiled/built react files when the
`NODE_ENV` env var is set to `production`.
These will be served from the `build` folder.


aidenkeating:
👀

priley86:
hey @david-martin - no objections from me given our conversation earlier today. 

adding @dlabaj @dgutride for review.

server.js
```diff
@@ -0,0 +1,27 @@
...
+    openshiftAPIUrl: "${process.env.OPENSHIFT_API_URL}",
+    accessToken: "${req.headers['x-forwarded-access-token']}"
+  }`)
+})
+
+if (process.env.NODE_ENV === 'production') {
+  // Serve any static files
+  app.use(express.static(path.join(__dirname, 'build')))
+  // Handle React routing, return all requests to React app
+  app.get('*', function (req, res) {
```
priley86: this would enable us to use `BrowserRouter` as opposed to `HashRouter` from a react-router perspective if desired. 👍

server.js
```diff
@@ -0,0 +1,27 @@
...
+
+// Dynamic configuration for openshift API calls
+app.get('/config.js', (req, res) => {
+  // The x-forwarded-access-token header will be forwarded
+  // by the oauth-proxy to this server.
+  // The below makes this token available to the client app
+  // for use in API calls to the configured OpenShift cluster.
+  res.send(`window.OPENSHIFT_CONFIG = {
+    openshiftAPIUrl: "${process.env.OPENSHIFT_API_URL}",
+    accessToken: "${req.headers['x-forwarded-access-token']}"
```
aidenkeating: Am I correct in saying both of these values have the possibility of being undefined? If they are this will return a string `"undefined"` instead of something like `null`. We should probably return `null`. wdyt?

david-martin:
Progressing changes in https://github.com/integr8ly/openshift-resource-parser/pull/1 for auth.
Will update this PR to use new version of openshift-resource-parser once merged & published

server.js
david-martin: For the oauth redirect flow, BrowserRouter would be preferrable.

david-martin:
@priley86 This is now ready for review/merge from my point of view.
I'm going to leave the update to the newer version of openshift-resource-parser (that has the oauth flow built in)  to a separate PR so as not to overload this one.

PR #8, opened 2018-08-31
moving docs from personal repo

finp:
moving docs from https://github.com/finp/in8-docs

PR #9, opened 2018-09-03
Docs update

finp:
replaces https://github.com/integr8ly/tutorial-web-app/pull/8

### DIFF
--- a/public/steelthreads/json/en/thread-0.json
+++ b/public/steelthreads/json/en/thread-0.json
@@ -1,149 +1,107 @@
 {
-  "id": 0,
-  "title": "Integrating two RESTful web services",
-  "descriptionDoc": "thread-0-description.adoc",
-  "prerequisites": [
-  "Access to an OpenShift instance with Integreatly components installed",
-  "A GitHub account configured with an OAuth App to use the Integreatly Launcher instance."
-  ],
-  "estimatedTime": 50,
-  "roles": [
-  "Operator",
-  "Developer"
-  ],
-  "applications": [
-  "Red Hat OpenShift Application Runtimes",
-  "Eclipse Che",
-  "Fuse Ignite",
-  "EnMasse"
-  ],
-  "tasks": [
-  {
-  "title": "Creating an EnMasse space",
-  "description": "EnMasse simplifies running messaging infrastructure for your organization. You use it to provide messaging services from a Node.js app to a Spring Boot app.",
-  "estimatedTime": 6,
-  "stepDoc": "setting-up-enmasse.adoc",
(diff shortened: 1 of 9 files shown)


public/steelthreads/json/en/thread-0.json
```diff
@@ -3,64 +3,147 @@
...
+  "estimatedTime": 5,
+  "stepDoc": "using-integration.adoc",
+  "stepDocSuccess": "thread-0-success.adoc"
+  }
   ]
-}
+  }
+  
+  // 20180903163703
+  // https://raw.githubusercontent.com/integr8ly/tutorial-web-app/8921fd954396fe26445553ada441b186d8fa7ca3/public/steelthreads/json/en/thread-0.json
```
priley86: so these comments are breaking my json validations... can you possible shore up this JSON and make sure it validates? You can use https://jsonlint.com/ for example.

priley86: not sure if you had some questions or just revising?

finp: updated to make valid, but really don't know about the fields in this file, so maybe I still don't have it right. 
eg I'd like to have mulitple  procedures in a task, could stepDoc be an array?

priley86: sure - we can make stepDoc an array if that is easier (and just render the array of docs). 

Regarding the verification, i don't think it is simple in Ascii format to add the icon spans (having been looking for a solution, but nothing seems simple). For now, I will just add them dynamically on the React side if OK. Will send you a commit shortly...

priley86: ok pushed a small change here for the alerts to try to meet the design styling... what do you think of this approach? Let me know if OK. https://github.com/priley86/integr8ly-prototype/commit/b090abfe387b7db8a5bd0a2039bc51d05ab674eb

priley86:
so I am thinking a structure like this will work much better for us since we still need formatting and links inside the verifications. Is there some way you can include the verifications within another ascii doc on your side easily? This is another path which I think will give you more flexibility with fewer steps than JSON...

https://github.com/priley86/integr8ly-prototype/commit/ee00eaffc9ebe5e22d5eacc2851adc63f30da1e1#diff-8dbd8cb7c1f7b2cd546ce1238ee20d19

stacymcauliffe:
@finp - one comments from me - we had a request to change the name of EnMasse to Red Hat AMQ Online (or just AMQ Online if you're allowed to use a short name in the docs).

finp:
closed in favor of https://github.com/integr8ly/tutorial-web-app/pull/17

PR #7, opened 2018-08-31
[WIP] Add oauth flow to the webapp

david-martin:
This adds a new route `/oauth/callback` to handle the oauth flow
redirect.
A simple openshift API call is included in the tutorial page to
retrieve service instance details.
This call will trigger an OAuth flow if the user isn't logged in.

TODO

* [x] Rebase after https://github.com/integr8ly/tutorial-web-app/pull/6 is merged
* [ ] Update the version of the openshift-resource-parser lib after https://github.com/integr8ly/openshift-resource-parser/pull/1 is merged & published
* [x] Update README steps for local dev for setting up OAuthClient in OpenShift
* [x] Verify the default config.js content works for local development (using mock data)
* [x] Verify local development can be configured to work with an openshift cluster (using real data for integration testing)

PR #11, opened 2018-09-04
Add env vars for mocked service urls

david-martin:
This change does a few things:

* Brings in the openshiftResourceParser library as a component instead
of an external dependency
* Modifies the parser to use a hardcoded list of services available i.e.
Fuse, Che, Launcher, EnMasse
* Sets the link for each of these services according to 4 new *build time* env vars (not runtime env vars):
  * REACT_APP_FUSE_URL
  * REACT_APP_LAUNCHER_URL
  * REACT_APP_CHE_URL
  * REACT_APP_ENMASSE_URL

A deploy of this can be see http://react-demo-app-webapp.cluster-lfa3xlh.opentry.me/ (This still uses the s2i build method with nginx for serving content)
The BuildConfig has these env vars set to the corresponding urls for those services in that cluster.

**NOTE** This change is not intended for production use. It is only for using configurable urls with a fixed mocked list of 4 services. Auto-discovery of services & their console urls will be added in a separate PR.


david-martin:
@johnfriz @aidenkeating would you mind taking a look (url & details in description)

priley86:
this is looking good to me... will test some locally here shortly.

src/components/openshiftResourceParser/clients/mock-provisioned-service-client.js
```diff
@@ -0,0 +1,68 @@
...
+        appDescription: 'A developer workspace server and cloud IDE.',
+        appLink: `${process.env.REACT_APP_CHE_URL}`
+      },
+      {
+        appName: 'EnMasse',
+        appDescription: 'Managed, self-service messaging on Kubernetes.',
+        appLink: `${process.env.REACT_APP_ENMASSE_URL}`
+      }
+    ];
+    debugger
```
dlabaj: remove debugger

src/pages/landing/landingPage.js
```diff
@@ -2,53 +2,36 @@ import * as React from 'react';
...
-  </div>
-);
+class LandingPage extends React.Component {
+  state = {
+    apps: []
+  }
+
+  componentDidMount() {
+    const parser = new OpenShiftResourceParser({ mockData: true });
+    parser.listProvisionedMWServices("eval")
```
priley86: per discussion w/ @dlabaj, would you be opposed to us porting `listProvisionedMWServices` into a Redux action in a future PR? The `OpenShiftResourceParser` and relevant services should be fine as is, we'd just like to hook into this `listProvisionedMWServices ` promise from a Redux action/reducer (and treat this like an API service). For now, this should work fine though! Redux is _optional_.

src/components/openshiftResourceParser/clients/mock-provisioned-service-client.js
david-martin: :+1:

src/pages/landing/landingPage.js
david-martin: No objections. (I'll have to do some reading on redux to fully grok what you are saying :) )

priley86:
@david-martin ok to merge? I think we are OK when you are.

src/pages/landing/landingPage.js
priley86: I think Redux definitely makes sense if we are also looking to use this list to drive the application launcher menu. This is a case where Redux's global state is perfect. Will follow up in future PR.

package.json
```diff
@@ -7,13 +7,13 @@
...
     "axios": "^0.18.0",
     "classnames": "^2.2.5",
+    "client-oauth2": "^4.2.1",
     "detect-browser": "^3.0.0",
     "express": "^4.16.3",
     "i18next": "^11.6.0",
     "i18next-xhr-backend": "^1.5.1",
     "js-cookie": "^2.2.0",
     "lodash.get": "^4.4.2",
-    "openshift-resource-parser": "^0.1.0",
```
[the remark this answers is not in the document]
david-martin: Can discuss further in https://github.com/integr8ly/openshift-resource-parser/issues/2 if you wish.
I wouldn't like to have it as a separate lib if it is going to become heavily tied to the tutorial web app (unnecessary churn of module build/release cycle).
Perhaps there's a piece of it (just the openshift api & auth stuff?) that could be pulled out, but I'd like to see that naturally happen over time, if that makes sense

david-martin:
Thanks @dlabaj  @priley86.
I'd like to create a PR that swaps out nginx for the node.js server (and BrowserRouter instead of HashRouter) next, which will include CI build for a docker image.

We can hold off on the auth/sso PR #7 a little longer as there's env config needed for it as part of that work.

PR #17, opened 2018-09-07
Docs  Update

finp:
updated docs as discussed:
docs/bin contains scripts to populate adoc files in public/


priley86:
@finp thanks very much for this PR. I think this will work, but need a bit of time to review it further. Initially looking at the new approach here, I think we are OK 👍

priley86:
@finp thanks again - i've cherry-picked these commits and made a few minor adjustments...
https://github.com/integr8ly/tutorial-web-app/commits/development

* tweaked "Verifications" header and alert div to be outside the partial (prefer to render that separately). Let me know if OK or you would like to adjust...
https://github.com/integr8ly/tutorial-web-app/commit/5557ccba6a47e7d63ff7df008e01ff9e75ce0c53

Otherwise - it works great! 👍

finp:
@priley86  looks good, closing this PR

PR #34, opened 2018-09-21
Create AMQ service

aidenkeating:
@david-martin Mind taking a look.

* Provision the AMQ ServiceInstance
* Watch the AMQ Statefulset and create a secret from its credentials once it's ready
* Create an annotation on the AMQ ServiceInstance with the service route

I need to add some docs to this stuff too.

## Verification

* Run `OPENSHIFT_HOST=<your-cluster> yarn start:dev`
* Login
* Ensure: AMQ is provisioned. A secret named amq-broker-credentials is created. You can access the AMQ dashboard from the webapp.


witmicko:
Dont have write access here so cant approve:
code looks good.
verified on my cluster along with installer amq pr.

david-martin:
@aidenkeating Seeing a minor issue with the AMQ link in the webapp gives an 'Application not available' error for some time (1min?) even though it says its provisioned
image

Eventually the management console is up, and works.
Is there anything obvious we need to do or aren't doing so that its only marked as provisioned when its running?

Other than that, verified. (Looking through the code now)

src/services/walkthroughServices.js
```diff
@@ -1,9 +1,9 @@
-import { list, create, watch, currentUser, OpenShiftWatchEvents } from './openshiftServices';
+import { list, create, remove, watch, update, currentUser, OpenShiftWatchEvents } from './openshiftServices';
 import { walkthroughTypes } from '../redux/constants';
 import { FULFILLED_ACTION } from '../redux/helpers';
-import { buildServiceInstanceResourceObj } from './serviceInstanceServices';
+import { buildServiceInstanceResourceObj, DEFAULT_SERVICES } from './serviceInstanceServices';
 
-const WALKTHROUGH_SERVICES = ['fuse', 'che', 'launcher', 'enmasse-standard'];
+const WALKTHROUGH_SERVICES = ['fuse', 'che', 'launcher', 'enmasse-standard', 'amq-broker-71-persistence'];
```
david-martin: should we reference these from the consts in serviceInstanceServices.js rather than having the names in 2 places?

e.g. `DEFAULT_SERVICES.ENMASSE`, `DEFAULT_SERVICES.AMQ` etc...

src/services/walkthroughServices.js
```diff
@@ -87,6 +142,27 @@ const handleServiceInstanceWatchEvents = (dispatch, event) => {
       type: FULFILLED_ACTION(walkthroughTypes.CREATE_WALKTHROUGH),
       payload: event.payload
     });
+
+    const dashboardUrl = 'integreatly/dashboard-url';
```
david-martin: might be worth a comment in here explaining why we're doing this i.e. because the template broker doesn't have dashboard url support.
Is there an issue in origin to request this feature?

src/services/walkthroughServices.js
```diff
@@ -112,6 +187,15 @@ const findOrCreateOpenshiftResource = (openshiftResourceDef, resToFind, compareF
     return Promise.resolve(foundResource);
   });
 
+const replaceOpenShiftResource = (resourceDef, resourceRes, replaceIfFn = (resObj => resObj.metadata.name === resourceRes.metadata.name)) =>
+  findOrCreateOpenshiftResource(resourceDef, resourceRes)
+    .then(resource => {
+      if (replaceIfFn(resource)) {
+        return remove(resourceDef, resourceRes).then(() => create(resourceDef, resourceRes));
```
david-martin: is this what the `oc replace` command does?
Is there actually not a replace api?

david-martin:
Some comments left. Nothing major though.

src/services/walkthroughServices.js
aidenkeating: Good catch, yep

src/services/walkthroughServices.js
aidenkeating: Will do, there is an issue around not being able to add a dashboardURL to a template.

https://github.com/openshift/origin/issues/21082

aidenkeating:
@david-martin The Provisioned status comes from looking at the ServiceInstance status. So it looks like the ServiceInstance is provisioned before the AMQ services are completely ready. I'll take a look into that.

dlabaj:
Looks good.  Just make those couple updates we talked about this morning.  Thanks.

aidenkeating:
@dlabaj Made the changes that we discussed.

@david-martin @witmicko I've made a change in 3a89b33 to store the AMQ Credentials for Walkthrough 1 in a Redux store. The idea is that when a user starts Walkthrough 1 we take the credentials from there and provide them as params to the Messaging app when provisioning. The Messaging app would then use these params in the template to populate the values it needs in its ConfigMap.

src/services/openshiftServices.js
```diff
@@ -27,8 +27,8 @@ const OpenShiftWatchEvents = Object.freeze({
 });
 
 class OpenShiftWatchEventListener {
-  _handler = () => {};
-  _errorHandler = () => {};
+  _handler = () => { };
```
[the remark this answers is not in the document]
aidenkeating: @dlabaj Didn't see a linting issue here, but there were a couple left from previous commits. I had to update a couple of snapshots to get it to pass. Would you mind taking another look?

PR #40, opened 2018-09-23
Updates from Development Branch

dlabaj:


src/styles/application/_copyField.scss
```diff
@@ -27,3 +24,32 @@
     }
   }
 }
+
+@media (min-width: 768px) {
```
mindreeper2420: Looks like this is repeated 3 times - only one with the 400px is needed.

PR #56, opened 2018-09-26
Walkthrough 1a

witmicko:


src/pages/landing/landingPage.js
```diff
@@ -46,7 +47,8 @@ LandingPage.defaultProps = {
 };
 
 const mapDispatchToProps = dispatch => ({
-  manageWalkthroughServices: () => manageMiddlewareServices(dispatch),
+  // manageWalkthroughServices: () => manageMiddlewareServices(dispatch),
+  manageWalkthroughServices: () => prepareWalkthroughNamespace(dispatch, walkthroughs.oneA),
```
aidenkeating: I don't think we want to prepare the walkthrough namespaces here. On the landing page we want to provision the ServiceInstance namespace. We probably want to provision the Walkthrough namespace when a user enters the walkthrough itself.

src/services/walktroughServices.js
```diff
@@ -0,0 +1,109 @@
...
+class NodeMessagingAppEnmassSiTransform {
+  isTransformable(siInfo) {
+    return siInfo.name === 'nodejs-messaging-work-queue-frontend';
+  }
+
+  transform(siInfo) {
+    const enmassCreds = {
+      MESSAGING_SERVICE_USER: 'username',
+      MESSAGING_SERVICE_PASSWORD: 'password'
+    };
```
aidenkeating: We'll probably need to set the internal service address here too to override the default URL.

src/pages/landing/landingPage.js
witmicko: +1, but create that after the user clicks on "Getting started" we don't want to setup the workspace if the user just wants to read the description before choosing the one walkthrough to do.

src/services/walktroughServices.js
JameelB: @aidenkeating Is that this URL here? `amqp://messaging.enmasse-eval.svc:5672?amqp.saslMechanisms=PLAIN`

src/services/walktroughServices.js
```diff
@@ -0,0 +1,109 @@
+import { create, watch, currentUser, OpenShiftWatchEvents } from './openshiftServices';
+import {
+  cleanUsername,
+  findOpenshiftResource,
+  buildServiceInstanceResourceObj,
+  findOrCreateOpenshiftResource
+} from './middlewareServices';
```
JameelB: Can we move `cleanUsername`, `findOpenshiftResource` and `findOrCreateOpenshiftResource` functions into commons since these can be reused here. I find it weird to be importing from `middlewareServices` here. wdyt?

I think the `buildServiceInstanceResourceObj` can be imported from `serviceInstanceHelpers`.

src/services/walktroughServices.js
```diff
@@ -0,0 +1,109 @@
...
+const walkthroughs = {
+  one: {
+    namespaceSuffix: 'walkthrough-one',
+    services: ['spring-boot-rest-http-crud', 'nodejs-messaging-work-queue-frontend'],
+    transforms: [new DefaultServiceInstanceTransform()]
+  },
+  oneA: {
+    namespaceSuffix: 'walkthrough-one-a',
+    services: ['spring-boot-rest-http-crud', 'nodejs-messaging-work-queue-frontend'],
+    transforms: [new NodeMessagingAppEnmassSiTransform(), new DefaultServiceInstanceTransform()]
```
JameelB: Is there a reason why we're including `DefaultServiceInstanceTransform` here?

src/services/walktroughServices.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+};
+
+const prepareWalkthroughNamespace = (dispatch, walkthrough) => {
+  currentUser().then(user => {
+    const userNamespace = `${cleanUsername(user.username)}-${walkthrough.namespaceSuffix}`;
+    const namespaceObj = osKinds.buildNamespaceObj(userNamespace);
+    findOpenshiftResource(osKinds.projects, namespaceObj)
```
JameelB: I think the function `findOrCreateOpenshiftResource` does the check below for you.

src/services/walktroughServices.js
```diff
@@ -0,0 +1,109 @@
+import { create, watch, currentUser, OpenShiftWatchEvents } from './openshiftServices';
```
JameelB: The filename for `walktroughServices.js` is missing an h => `walkthroughServices.js`

src/services/walktroughServices.js
witmicko: hey, it transforms the spboot app, at the moment only node app needs parameters
I was going to include defaulttransform as a fallback, so we dont have to always add it as last, but decided against it as it might cause confusion.

src/services/walkthroughServices.js
```diff
@@ -14,12 +14,14 @@ class NodeMessagingAppEnmassSiTransform {
   }
 
   transform(siInfo) {
-    const enmassCreds = {
+    const placeholderParameters = {
       MESSAGING_SERVICE_USER: 'username',
-      MESSAGING_SERVICE_PASSWORD: 'password'
+      MESSAGING_SERVICE_PASSWORD: 'password',
+      MESSAGING_SERVICE_HOST: 'messaging.enmasse-eval.svc',
```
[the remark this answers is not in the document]
aidenkeating: Yep, all these will need to be retrieved from EnMasse as they're all namespace specific. We can build this up on the client side if we want though as we know the name of the namespace (since we create it) and we know the service will be called messaging. So basically replace `enmasse-eval` with the known namespace name.

src/pages/landing/landingPage.js
aidenkeating: Not sure if you've resolved this already. But the approach I took is to start creating the services when the user enters the task pages instead of the tutorial pages. This ensures they've entered the steps of the tutorial.

src/services/walkthroughServices.js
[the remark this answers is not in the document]
witmicko: yup, I was waiting for that to be ready, hence `placeholderParams` as a name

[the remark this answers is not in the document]
JameelB: I've added the url now as well: https://github.com/integr8ly/tutorial-web-app/pull/60/files#diff-ed7417901bdafe31ec92bf19d0ea8a42R53

dlabaj:
@witmicko Looks like there is merge conflicts with this PR now.

src/pages/landing/landingPage.js
dlabaj: Agree with @aidenkeating we should prepare the walkthrough when the user enters the tutorial page.

src/common/openshiftKinds.js
```diff
@@ -0,0 +1,62 @@
+const namespaceRequestDef = {
```
dlabaj: Not sure what is meant by Kinds?  Any thought on a different name ... maybe openshiftHelpers.js?

src/services/middlewareServices.js
```diff
@@ -1,16 +1,13 @@
...
-  namespaceRequestDef,
-  namespaceDef,
-  namespaceRequestResource,
-  statefulSetDef,
-  routeDef,
-  secretDef
+  findOpenshiftResource,
+  findOrCreateOpenshiftResource
 } from '../common/openshiftHelpers';
+import * as osKinds from '../common/openshiftKinds';
```
dlabaj: Can we just import what is needed instead of all the exports?  What do you think?

src/common/openshiftKinds.js
witmicko: I can move it to helpers, but wanted to have k8s/openshift types in a separate file, as they are mostly simple objects

witmicko:
@dlabaj this is ready for review if you have some time.

src/common/serviceInstanceHelpers.js
```diff
@@ -13,7 +17,7 @@ class DefaultServiceInstanceTransform {
       },
       spec: {
         clusterServiceClassExternalName: siInfo.name,
-        clusterServicePlanExternalName: `default-${siInfo.name}`
+        clusterServicePlanExternalName: `default`
```
JameelB: @witmicko Can we revert the change to this? This is causing the provison of che, launcher and fuse to fail because the plan name we currently have to the broker is in this format: `default-${siInfo.name}`

witmicko:
@JameelB @aidenkeating addressed comments and squashed, that had dismissed your reviews

PR #63, opened 2018-09-28
Feed from testing part 3 updates, and bug fixes.

dlabaj:
* State updates for storing progress for users. 
* Fix for not being able to run locally.


src/services/userServices.js
```diff
@@ -186,4 +186,31 @@ const storeData = (data, remove = false, config = { extend: true }) =>
...
+  window.localStorage.setItem(
+    `userProgress-${window.localStorage.getItem('currentUserName')}`,
+    JSON.stringify(progress)
+  );
+  // window.localStorage.setItem(`threadState-${thread}`, JSON.stringify(threadState));
+  return progress;
+};
+
+const getProgress = () =>
+  // const progress = {};
```
mfrances17: Maybe add a TBD with why it's commented out for now, or just remove

dlabaj: Acutally this should be deleted good catch

PR #64, opened 2018-09-28
update gitignore to remove yarn.lock

mindreeper2420:


.gitignore
```diff
@@ -12,6 +12,7 @@ src/**/*.css
 
 # dependencies
 /node_modules
+yarn.lock
```
mfrances17: this is a total nit but i would put it under misc since it's not really a dependency... will work the same regardless

mindreeper2420: agree - fixed

Issue #14, opened 2018-09-06
Automatically set the external OpenShift Host on server startup


david-martin:
When running inside an OpenShift Cluster, the `/.well-known/oauth-authorization-server` path can be called to get cluster oauth information, which inlcludes the external cluster url. e.g.

```
curl -k https://kubernetes.default/.well-known/oauth-authorization-server
{
  "issuer": "https://mycluster.example.com",
  "authorization_endpoint": "https://mycluster.example.com/oauth/authorize",
  "token_endpoint": "https://mycluster.example.com/oauth/token",
  "scopes_supported": [
    "user:check-access",
    "user:full",
    "user:info",
    "user:list-projects",
    "user:list-scoped-projects"
  ],
  "response_types_supported": [
    "code",
    "token"
  ],
  "grant_types_supported": [
    "authorization_code",
    "implicit"
  ],
  "code_challenge_methods_supported": [
    "plain",
    "S256"
  ]
}
```

This could be done on startup of the server if it knows its running in an OpenShift cluster, avoiding the need to manually set the `OPENSHIFT_HOST` env var in the DeploymentConfig

PR #67, opened 2018-10-02
add ability to determine openshift host dynamically

evanshortiss:
See issue #14 for more details. I tested this by hitting the `/config.js` endpoint after deploying on OpenShift and verifying hosts were populated correctly. 

I wasn't able to get the full auth flow to work so couldn't test this end to end. The auth flow wouldn't work locally with the development branch either - this might be expected during local dev. Happy to try a more complete test if anyone knows what auth issue I might be seeing.

screen shot 2018-09-28 at 12 48 38


Dockerfile
```diff
@@ -10,6 +10,8 @@ COPY . ./
 
 USER root
 
+RUN chmod g+w yarn.lock
```
evanshortiss: Not 100% sure why, but I needed this to get around permissions with the lockfile.

server.js
```diff
@@ -1,12 +1,53 @@
...
+const https = require('https')
+const port = process.env.PORT || 5001
+const isProduction = process.env.NODE_ENV === 'production'
 
-const app = express();
-const port = process.env.PORT || 5001;
+async function requestHostFromKubernetes() {
+  const response = await axios({
+    url: 'https://kubernetes.default/.well-known/oauth-authorization-server',
+    httpsAgent: new https.Agent({ rejectUnauthorized: false })
```
evanshortiss: This ignores SSL errors. I was going to use `isProduction` to determine this, i.e in production we'd expect a valid SSL cert (`rejectUnauthorized: true`) but in dev we can ignore the cert errors (`rejectUnauthorized: false`). That doesn't work though since `isProduction` _has_ to be true to expose the assets both during development and production using `express.static`.

Could possibly add a new env variable to manage this?

server.js
```diff
@@ -1,12 +1,53 @@
-const express = require('express');
-const path = require('path');
+const express = require('express')
```
dlabaj: Missing ; at the end of statements.

server.js
dlabaj: If you need to add a new env variable to support this you can go ahead and add it to the .env files.

server.js
```diff
@@ -1,12 +1,53 @@
...
+  } else {
+    return response.data.issuer;
+  }
+}
+
+async function getOpenShiftHost() {
+  let { OPENSHIFT_HOST } = process.env;
+
+  if (!OPENSHIFT_HOST) {
+    console.log('OPENSHIFT_HOST environment variable not set. attempting to fetch host programmatically')
```
dlabaj: ; Missing at end of line.

server.js
```diff
@@ -77,17 +118,17 @@ app.get('/config.js', (req, res) => {
...
+      accessTokenUri: 'https://${OPENSHIFT_HOST}/oauth/token',
+      authorizationUri: 'https://${OPENSHIFT_HOST}/oauth/authorize',
       redirectUri: '${redirectHost}/oauth/callback',
       scopes: ['user:full'],
-      masterUri: 'https://${process.env.OPENSHIFT_HOST}',
-      wssMasterUri: 'wss://${process.env.OPENSHIFT_HOST}'
-    };`);
+      masterUri: 'https://${OPENSHIFT_HOST}',
+      wssMasterUri: 'wss://${OPENSHIFT_HOST}'
+    };`)
```
dlabaj: ; Missing at end of line.

Dockerfile
dlabaj: We where getting an error before so we removed the yarn.lock file.  This looks like this is a better solution.  If this works then let's go ahead and add the yarn.lock file back.   I'm not sure why it's needed either.

evanshortiss:
@dlabaj updated with requested changes 👍 

It might be worth adding `husky` to run prettier in a `precommit` hook to automate the formatting.

dlabaj:
@evanshortiss Adding husky to run prettier sounds like a great idea.

PR #70, opened 2018-10-03
Pull together Redux store and docs for Walkthrough One

aidenkeating:


src/pages/tutorial/task/task.js
```diff
@@ -94,6 +96,48 @@ class TaskPage extends React.Component {
     setProgress(progress);
   };
 
+  areDocLinksReady = () => {
```
aidenkeating: @dlabaj I've added this in as a temporary fix for the Asciidoc not re-rendering when I update the attributes. I'll look into why the Asciidoc isn't re-rendering.

src/services/walkthroughServices.js
```diff
@@ -25,24 +27,26 @@ import {
...
     watchers: [
       {
         resource: namespace => routeDef(namespace),
-        handlerFn: handleWalkthroughOneRoutes
+        handlerFn: handleWalkthroughOneRoutes.bind(null, 'walkthrough-one')
       }
     ]
   },
   oneA: {
+    id: WALKTHROUGH_IDS.ONEA,
```
witmicko: nit: `ONE_A` ?

aidenkeating:
7577c90 is an attempt to make the asciidoc component reactive. Instead of storing the compiled html we just store whatever is returned from the adoc or template endpoint and compile it on the render call. Doesn't seem to have any performance issues and it means I can remove the hacky "loading" state I had in before.

aidenkeating:
@dlabaj Mind taking a look?

aidenkeating:
@dlabaj Think this is ready to merge

PR #71, opened 2018-10-03
WIP: Docs refactor

finp:
The goal of this PR is to enable:
* writers to write content using mod-docs standards, methods and protocols
* automatically generate the fragments/json required by app


finp:
@aidenkeating  still working on this, but would appreciate any feedback

aidenkeating:
👀

docs/modules/ROOT/pages/master-1.adoc
```diff
@@ -0,0 +1,63 @@
...
+
+++++
+<img src="/images/st0.png" class="img-responsive" alt="integration"> 
+++++
+
+// end::intro[]
+endif::location[]
+
+
+include::{partialsdir}/enmasse.adoc[leveloffset=1]
```
aidenkeating: For AMQ we don't need any setup in the AMQ console. So I'd guess we'd only need `connections`, `creating-fuse-integration` and `using-integration` here?

I'm also wondering, for `connections`. In Walkthrough One this'll be creating an `AMQ Connection` from Fuse. In Walkthrough One-A it'll be creating an `AMQP Connection` from Fuse.

Their wizards are different. So the steps won't match up between the two walkthroughs. Should be have separate `connection` files for each of the walkthroughs?

public/steelthreads/json/en/thread-2.json
```diff
@@ -0,0 +1,53 @@
...
+      "title": "Creating an integration",
+      "estimatedTime": 5,
+      "steps": [
+        {
+          "stepDoc": "creating-fuse-integration.adoc",
+          "infoVerifications": [
+            "creating-fuse-integration-verification.adoc"
+          ],
+          "attributes": {
+            "fuse-url": "https://cluster-lfa3xlh.opentry.me"
```
aidenkeating: Which of these `fuse-url`s should take precedence? The top guy or this guy?

finp: I'll remove 'this guy'

docs/modules/ROOT/pages/_partials/task-creating-fuse-integration.adoc
```diff
@@ -0,0 +1,65 @@
...
+
+. Select *Integrations* from the left hand menu.
+
+. Select the *Create Integration* button to start the *Create Integration* wizard.
+
+. Choose `Walkthrough One Messaging App` as the *Start Connection*.
+
+. When prompted to *Choose an Action*, select *Subscribe for messages*.
+
+. When prompted for a *Destination Name*, enter `work-queue/requests`.
```
aidenkeating: Nothing to do with this PR, but is there anything in Asciidoc that renders an easily copyable field? So they'd just have to click a button or something to save to their clipboard instead of copy the `work-queue/requests` part themselves?

PR #77, opened 2018-10-05
Removed hard coded days & create threads.json

dlabaj:


aidenkeating:
:eyes:

public/steelthreads/json/en/threads.json
```diff
@@ -0,0 +1,3 @@
+{
```
aidenkeating: @dlabaj @finp How will this file be used?

finp: don't know, but it is the array of walkthrough IDs that currently exist

dlabaj: I'm going to use it get the list of walk throughs that are available.  Since we are pulling the files from the server without a REST interface it was the solution I thought of.  Does this sound reasonable to you two?  @finp @aidenkeating

aidenkeating: @dlabaj Is this the list of walkthroughs on the landing/home page? If so, should the title and description be stored along with these IDs to populate the UI?

finp: +1 to the idea of storing titles and 'one liners' in this file. Currently, these are hard coded in a js file, and we're trying to single source those strings from docs

aidenkeating: @dlabaj @finp So those one liners would come from the asciidoc, similar to default attributes? If so, I'll create a card and we can address it in a separate PR.

dlabaj: Going to remove this code for now and resubmit it as a separate issue.  Will just hand code the threads.json file and come back to this after we get the bugs needed for the demo fixed.

PR #81, opened 2018-10-05
Fix status buttons

mfrances17:
Changes/fixes:

Removed hard-coded radio buttons on bottom of panel, they're now auto-generated and named according to the actual steps.

Next button inactive until all checkboxes are enabled (aka completed), then enabled/color change

Handles info steps as well as success steps (fixed exception err)

Footer 'success' buttons now toggle and are coordinated with checkboxes, so when you enable/disable a checkbox, the icon changes to a checkmark and is colored green/circle icon and is colored blue

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,10 +5,10 @@ src/**/*.css
 
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
+.vscode/settings.json
+.vscode/tasks.json
+.vscode/launch.json
+.vscode/extensions.json
 
 # dependencies
 /node_modules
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -8,10 +8,10 @@ src/**/*.css
 .idea
 # IDE - VSCode
 .vscode/*
-!.vscode/settings.json
-!.vscode/tasks.json
-!.vscode/launch.json
-!.vscode/extensions.json
+.vscode/settings.json
```
dlabaj: Not sure we should be ignoring these since they are setup for this project.  If you made project changes, would recommend unchecking them out and setup as user settings.

src/pages/tutorial/task/task.js
```diff
@@ -246,12 +254,58 @@ class TaskPage extends React.Component {
...
-                      <span className="integr8ly-module-column--footer_status-step">1.2</span>
-                      <Icon type="fa" name="circle-o" />
-                      <span className="integr8ly-module-column--footer_status-step">1.3</span>
+                      {threadTask.steps.map((step, l) => (
+                        <React.Fragment key={l}>
+                          {step.infoVerifications &&
+                            step.infoVerifications.map(() => (
+                              <Icon
+                                className={
+                                  step.infoVerifications && verifications[step.infoVerifications[0]]
```
dlabaj: the arrays for infoverifications and successverifications should be changed to scalars.

PR #82, opened 2018-10-06
Added support for Windows.

dlabaj:
Added support for local development for Windows.


dlabaj:
Will add in future when things settle down.

PR #93, opened 2018-10-09
Include the TCP and AMQP credentials in the middlwareReducer

aidenkeating:
Currently we only store the TCP AMQ broker, however we require the
AMQP broker service url also. This change includes the AMQP and TCP
broker URLs.


aidenkeating:
Image available at `docker.io/aidenkeatingrht/tutorial-web-app`

PR #102, opened 2018-10-10
enable travis for pull requests to master

pb82:
For some reason travis builds are not triggered for PRs. Trying to enable them.


mindreeper2420:
Is there a specific need for it to build with Travis? I'm not sure I understand the benefit to having Travis with this, at the moment.

pb82:
@mindreeper2420 It runs the tests for every pull request and we can see the status in Github. So if e.g. someone forgets to update the snapshots this would catch it.

mindreeper2420:
Ok, sounds good. That'd be a great thing to have in the PRs.

pb82:
@mindreeper2420 Automatic tests should be working now. You can see the details by clicking on `Show all checks -> Details`. I also added the yarn.lock file to the repo as suggested here: https://yarnpkg.com/lang/en/docs/yarn-lock/#toc-check-into-source-control

Does that look ok to you?

dlabaj:
@pb82 @mindreeper2420 We removed it because we where having issues on the build environments they kept getting a permission error.  If that's fix then we should go ahead and add it back.

dlabaj:
@pb82 It was an openshift issue we where seeing when trying to build the application.  It was fixed in this PR https://github.com/integr8ly/tutorial-web-app/pull/67 .  It should be fine to check it in now.

PR #107, opened 2018-10-10
fix estimatedTime value

odra:
fixes the estimatedTime value for each thread

### DIFF
--- a/public/steelthreads/json/en/thread-0.json
+++ b/public/steelthreads/json/en/thread-0.json
@@ -6,7 +6,7 @@
     "Access to an OpenShift instance with Integreatly components installed",
     "A GitHub account"
   ],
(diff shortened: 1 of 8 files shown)


pb82:
@odra What do you think about removing the total estimated time property and instead just summarize the estimated time programatically here: https://github.com/integr8ly/tutorial-web-app/blob/master/src/pages/tutorial/tutorial.js#L159

odra:
I thought the same but that would require more time to do (at least for a react wannabe like me) because we would have to change how the index page retrieves the theads json file (it retrieves a general threads json file which contains only one time field) - maybe we should move to one big json file in the future.

I was thinking in just fixing the value sum for now and maybe creating a new card to fix the json structure stuff.

PR #109, opened 2018-10-10
Add status unavailable

JameelB:
## Description
Add the status unavailable to capture states when the service provision fails and when the service gets deleted.

image

### DIFF
--- a/src/components/installedAppsView/InstalledAppsView.js
+++ b/src/components/installedAppsView/InstalledAppsView.js
@@ -28,7 +28,20 @@ class InstalledAppsView extends React.Component {
         <Icon type="pf" name="on-running" /> &nbsp;Ready for use
       </div>
     );
+    const unavailableStatus = (
+      <div className="state-unavailable">
+        <Icon type="pf" name="error-circle-o" /> &nbsp;Unavailable
+      </div>
+    );
+
+    if (app.metadata && app.metadata.deletionTimestamp) {
+      return unavailableStatus;
+    }
+
     if (app.status && app.status.conditions && app.status.conditions[0]) {
+      if (app.status.provisionStatus === 'NotProvisioned') {
+        return unavailableStatus;
+      }
       return app.status.conditions[0].status === 'True' ? readyStatus : provisioningStatus;
     }
     return provisioningStatus;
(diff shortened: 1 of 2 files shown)


JameelB:
@aidenkeating @jasonmadigan  Tests are not passing in master, will fix in a separate PR.

src/components/installedAppsView/InstalledAppsView.js
```diff
@@ -17,11 +17,31 @@ class InstalledAppsView extends React.Component {
...
+    let statusStyle;
+    let provisioningStatus = app.status.provisionStatus || 'Provisioning';
+
+    statusIcon = provisioningStatus === 'Provisioned' ? 'pficon-ok' : 'spinner spinner-xs spinner-inline';
+    statusStyle = provisioningStatus === 'Provisioned' ? { color: '#3f9c35' } : { color: '#4d5258' };
+
+    if (provisioningStatus === 'NotProvisioned') {
+      provisioningStatus = 'Provision Failed';
+      statusIcon = 'pficon-error-circle-o';
+      statusStyle = { color: '#cc0000' };
```
JameelB: I tried using the patternfly color palette here like `{ color: '$pf-red' } but I couldn't get it working.

mindreeper2420: The correct variable is `$color-pf-red-100`.

catrobson:
@JameelB @aidenkeating - Can you clarify what all the states are that we have available at this point so we can update the design if needed?

JameelB:
@mindreeper2420 Looks like the addition of `Ready for Use` and `Provisioning` was already done here https://github.com/integr8ly/tutorial-web-app/pull/110. I've added the status `Unavailable` here instead to capture the states of when the service fails provision and when it gets deleted.

@catrobson At the moment, we currently have `Ready for Use`, `Provisioning` and `Unavailable`. I did see that there was a state `Restarting` in the design. I'm not sure if we can have this as we're currently watching the service and as far as I know we can't restart a service on OpenShift.

aidenkeating:
@JameelB @catrobson The states that are possible for the services are:

* `Provisioning` - This is implemented in #110 
* `Provisioned/Ready for Use` - This is implemented in #110 
* `Error/Unavailable` - This is implemented in this PR

`Restarting` is not a possible state. The OpenShift ServiceInstance that backs this service information can't be restarted, only created and removed.

PR #128, opened 2018-10-12
Update walkthrough 1A diagram and resources

JameelB:
The following changes includes:
- Fix broken image for the `Walkthrough 1A` architecture diagram.
- Move the walkthrough resource definitions to the `thread-{walkthrough}.json` files and create a walkthrough resources component to render the list.
- Add Enmasse resource to the `Walkthrough 1A` resource list.

image


public/steelthreads/json/en/thread-1A.json
```diff
@@ -119,5 +119,54 @@
...
+    {
+      "title": "Red Hat AMQ Online",
+      "links": [
+        {
+          "name": "Open Console",
+          "url": "https://url/"
+        },
+        {
+          "name": "Red Hat AMQ Online Overview",
+          "url": "https://www.redhat.com/en/explore/amq-online"
```
JameelB: Is there any links in particular that we need to add here? @jasonmadigan @aidenkeating

pb82:
@JameelB I can't see EnMasse in the resources list for 1A. Is is supposed to be there?

src/components/walkthroughResources/walkthroughResources.js
```diff
@@ -0,0 +1,43 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const WalkthroughResources = props => {
+  const { resources } = props;
+  let resourceList;
```
pb82: It might be better to initialize `resourceList` to `null` instead of `undefined`. React can't render undefined, but null is fine: https://github.com/facebook/react/issues/3738

witmicko: TIL, good point.

PR #129, opened 2018-10-12
Dashboard and task numbering updates.

dlabaj:


src/components/installedAppsView/InstalledAppsView.js
```diff
@@ -64,7 +64,10 @@ class InstalledAppsView extends React.Component {
         key={`openshift_console_${index}`}
         value={index}
       >
-        OpenShift Console
+        <p>Red Hat OpenShift</p>
+        <div className="state-ready">
+          <Icon type="pf" name="on-running" /> &nbsp;Ready for use
```
mindreeper2420: I would add a style attribute to the icon (something like `margin-right: 5px` instead of using a non-breaking space to add distance between the icon and the text.

dlabaj: Updated to use the css instead of &nbsp.

PR #131, opened 2018-10-14
Rolled back redux and fixed no loading between tasks in mock mode.

dlabaj:
- Rolled back redux due to create store error being introduced in new build.
- Fixed bug with tasks getting stuck on loading in mock server mode.


aidenkeating:
@dlabaj Verified the mock mode loading works. Just provisioning my cluster at the moment to confirm normal-use loading screen works correctly. Will approve once confirmed.

PR #134, opened 2018-10-15
Dashboard url fallback

pb82:
When `status.dashboardUrl` does not exist on a service instance we use `annotations['integreatly/dashboard']` as a fallback. But the annotations object might not exist when, for example, a service failed to provision. This change provides a fallback to the fallback by checking for existence first.

Also fixes some react warnings in the browser console.


aidenkeating:
:eyes:

Issue #126, opened 2018-10-12
Yarn lock file causing openshift permission failures


dlabaj:
There seems to be a permission issue with the yarn.lock file that's causing it to fail to run on openshift.  We will need to figure out what's causing this so we can get the yarn.lock file checked in again.

PR #137, opened 2018-10-16
another attempt to add the lockfile

pb82:
testing if the image can be built with the lockfile present


pb82:
@dlabaj do you want to give this another try?

dlabaj:
Fixed issue #126

PR #138, opened 2018-10-16
Set open console urls

JameelB:
## Description
- Update the `threads-{walkthrough}.json` files to include the service name for each resource. 
- Retrieve each walkthrough resource console url from the middlewareServices list in the store using the service name as an identifier and set the url of the `Open Console` link.
- Update snapshot to use `connect`


pb82:
👀

src/components/walkthroughResources/walkthroughResources.js
```diff
@@ -1,43 +1,94 @@
...
+      return resources.map(resource => {
+        let url = '';
+        if (resource.serviceName === 'openshift') {
+          url = `${window.OPENSHIFT_CONFIG.masterUri}/console`;
+        } else {
+          url = getDashboardUrl(middlewareServices.data[resource.serviceName]);
+        }
+
+        resource.links.forEach(link => {
+          if (link.name === 'Open Console') {
```
pb82: Is there ever a chance that the UI will be translated? Because then this would no longer work. Any other way to identify the link?

JameelB: good catch. Atm there's none, but I can probably put an id to along the lines of `consoleURL`

src/components/walkthroughResources/walkthroughResources.js
```diff
@@ -1,43 +1,94 @@
...
+    return null;
+  }
+
+  buildResourceList() {
+    const resources = this.mapServiceLinks();
+    let resourceList = null;
+    if (resources && resources.length !== 0) {
+      resourceList = resources.map(resource => {
+        let resourceLinks;
+        if (resource) {
```
pb82: Is this check really necessary? Can the `resources` array contain undefined items?

JameelB:
@pb82 updated, mind taking another look?

PR #141, opened 2018-10-16
Fix broken stylelint tests

mindreeper2420:


src/components/tutorialDashboard/tutorialDashboard.js
```diff
@@ -41,17 +41,17 @@ const TutorialDashboard = props => {
         >
           <p>{walkthrough.descriptionDoc}</p>
 
-          <div className="walkthrough-labels">
+          <div className=" integr8ly-walkthrough-labels">
```
mfrances17: extra leading space

mindreeper2420: fixed

PR #143, opened 2018-10-16
Return empty object for walkthrough docs in mock mode.

dlabaj:


dlabaj:
Probably want a better solution long term then returning an empty object.

PR #146, opened 2018-10-17
Add verification negative content for thread-2

mfrances17:
thread-2 scenario content was added recently but without any 'verification no' content, which led to no content drop-downs on selection of No. Added this in so the thread-2 scenario will be consistent with the other two scenarios.

This will be doc responsibility going forward for any new scenarios that have verification(s). Each verification must have a corresponding .VerificationNo section in the _partials adoc and a infoVerificationsNo key/value in the thread-#.json file.


david-martin:
:eyes:

mfrances17:
These are placeholders in order to support the current approved design. I don't work for documentation. It was assumed that a writer would add the information once the structure was in place.

stacymcauliffe:
For the verification steps, we really need to include a "no" state, even if it has to be fairly generic at this stage. Otherwise, we put users in a position where they have no way to move forward if they don't see what they're supposed to see. 

For now, following David's comment, a reasonable generic message would be, "Verify that you followed each step in the procedure above with attention to names, ids, and buttons. If you're still having issues, contact your administrator."

Originally, we had hoped there would be some way to automate verification on the back end, in which case we'd be able to display more specific error states and troubleshooting. Moving forward, it would make sense to explore the potential for automated verification and/or creating some kind of troubleshooting guide or forum to help users address common error states.

Let me know if this is helpful @dlabaj @mfrances17 @david-martin @finp

david-martin:
Thanks for clarification @stacymcauliffe .

I think this is now superceded by https://github.com/integr8ly/tutorial-web-app/pull/158 (<-- I'll review that now)

finp:
@stacymcauliffe thanks for the background, closing this PR in favor of  #158

PR #148, opened 2018-10-17
Add wt2.png diagram & update timings for steps

david-martin:
Walkthrough Overview should look as below

image


aidenkeating:
:eyes:

david-martin:
> The image is larger than the others, is then intentional/does it matter?

I didn't notice.
Looks like it might be based on the diagram content.

The only reason I can think of why it might matter is that the image expands to its maximum size as you resize the browser.
There is no min or max width.
So theoretically 2 different users could view the diagram in different extreme ways e.g.

image

image

cc @dlabaj or ux input

david-martin:
Going to merge this as is.
If there are any ux concerns, lets discuss and create a follow up pr if needed

Issue #145, opened 2018-10-16
Navigation/state issue with Next button


mfrances17:
If you click **Yes** and then click **Next**, then click **Back**, your Yes selection is remembered but the Next button is disabled so you can't go to the next panel. You have to click the Yes button again to proceed to the next panel.

PR #162, opened 2018-10-18
fix state issue of the next button in each walkthrough tasks

JameelB:
Add verification progress from the store to the initial state when the task loads to fix the state issue with the `Next` button being disabled when the steps are already verified.

Resolves https://github.com/integr8ly/tutorial-web-app/issues/145


pb82:
👀

PR #167, opened 2018-10-19
Update numbering

finp:
* manually add numbering to adoc files
* make all step headings visible


mfrances17:
Looks good so far. For the interim solution, we'll want to get the resulting output files in as part of this PR otherwise it's not worth pushing up.

finp:
@mfrances17 
I rebased on https://github.com/integr8ly/tutorial-web-app/tree/fixBottomNumbering to test this,
and noticed a problem navigating thru WT 1A, clicking 'Next' doesn't work, but I could set the URL manually to make the app work

finp:
@mfrances17 
I rebased on https://github.com/integr8ly/tutorial-web-app/tree/fixBottomNumbering to test this,
and noticed a problem navigating thru WT 1A, clicking 'Next' doesn't work, but I could set the URL manually to make the app work

mfrances17:
> @mfrances17
> I rebased on https://github.com/integr8ly/tutorial-web-app/tree/fixBottomNumbering to test this,
> and noticed a problem navigating thru WT 1A, clicking 'Next' doesn't work, but I could set the URL manually to make the app work

Wondering if the naming convention (introducing the period in the filename instead of dash) you used for the 1A adocs might have been the culprit. I see that you just renamed them, if that doesn't fix it, we can log a defect and investigate.

PR #169, opened 2018-10-22
add error screen

witmicko:
added error screen

verification:

remove either node-messaging or spboot template (back it up to file first) 
click into walkthrough
it will error out (missing template)

@aidenkeating @dlabaj 

published to quay.io/witmicko/tutorial-web-app:master
updated view:
image


mindreeper2420:
@witmicko the "How do you want to proceed?" text should be on a new line.

witmicko:
@mindreeper2420 thank you for review, updated now.

src/styles/application/_errorScreen.scss
```diff
@@ -0,0 +1,42 @@
...
+    left: 0;
+    height: 100%;
+    width: 100%;
+    background-size: cover;
+    background-color: $pf-color-blue-600;
+    background-image: url("../../img/provisioning-bg.png");
+    z-index: 5;
+  }
+
+  &-error-img {
```
mindreeper2420: Percentages should not be used to set margins. Using the proper template styling, this is not needed (see `loading.scss`.

src/styles/application/_errorScreen.scss
```diff
@@ -0,0 +1,42 @@
...
+    margin-bottom: -20%;
+  }
+
+  &-logo {
+    background-size: cover;
+    height: 36px;
+    width: 113px;
+    position: absolute;
+    top: 85px;
+    right: 50px;
```
mindreeper2420: This should be `right: 25px;` in order to match the positioning on the loading screens - all placements of the Red Hat logo should be the same.

src/components/errorScreen/errorScreen.js
```diff
@@ -0,0 +1,63 @@
...
+  };
+
+  render() {
+    if (!this.state.show) {
+      return '';
+    }
+
+    return (
+      <React.Fragment>
+        <Grid.Row className="integr8ly-errorscreen-backdrop">
```
mindreeper2420: This should match the template from `loadingScreen.js`, where the Bootstrap grid is replaced with `display: flex`. This allows for a natural layout with minimal forced positioning.

package.json
```diff
@@ -14,6 +14,7 @@
     "i18next-xhr-backend": "^1.5.1",
     "js-cookie": "^2.2.0",
     "lodash.get": "^4.4.2",
+    "lodash": "^4.17.10",
```
dlabaj: Do we really need to bring in lodash?

src/common/serviceInstanceHelpers.js
```diff
@@ -148,6 +150,20 @@ const handleWalkthroughOneRoutes = (namespacePrefix, dispatch, event) => {
   }
 };
 
+const handleServiceInstancesProvision = (namespacePrefix, dispatch, event) => {
+  const si = event.payload;
+  const lastCondition = _.last(si.status.conditions);
```
dlabaj: Prefer if we didn't use lodash and just got the last element using plain javascript.

src/common/serviceInstanceHelpers.js
```diff
@@ -148,6 +150,20 @@ const handleWalkthroughOneRoutes = (namespacePrefix, dispatch, event) => {
   }
 };
 
+const handleServiceInstancesProvision = (namespacePrefix, dispatch, event) => {
+  const si = event.payload;
```
dlabaj: Can we call this serviceInstance instead of si

src/common/serviceInstanceHelpers.js
dlabaj: Same comment as earlier in the review about removing lodash.

package.json
witmicko: @dlabaj personal preference for utility funcs, I ll remove it.

src/styles/application/_errorScreen.scss
```diff
@@ -0,0 +1,49 @@
...
+  &-backdrop {
+    position: fixed;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: 100%;
+    background-size: cover;
+    background-color: $pf-color-blue-600;
+    background-image: url("../../img/provisioning-bg.png");
+    z-index: -5;
```
mindreeper2420: A negative z-index should be avoided. From what I can see, if this error page is reused inside of the content area, it would collide with the fixed elements (thinking ahead to the dynamic/framed approach).

If the masthead is causing an issue, you can change the top positioning of the background to `40px`. (`top: 40px;`). Other adjustments can be made at a later date to deal with the problematic masthead positioning.

PR #176, opened 2018-10-25
WIP: Allow adding of custom URLs to the landing page

aidenkeating:
@david-martin Would you mind taking a look? The idea is that a configmap will get mounted at a location and the web app will pick up on this. I've been running it locally with the following command:
```
OPENSHIFT_HOST=master.akeating.openshiftworkshop.com SERVER_EXTRA_CONFIG_FILE=/tmp/meh.json yarn start:dev
```


aidenkeating:
@david-martin 
```
{
  "services": [
    {
      "name": "My Service",
      "url": "https://google-{{ username }}.com"
    },
    {
      "name": "My Other Service",
      "url": "https://google.com"
    }
  ]
}
```

aidenkeating:
@mindreeper2420 It's a dependency that's used on the server-side. There was a requirement to add custom URLs that can have the username included. Handlebars allows us to extend this functionality later on instead of doing something custom.

mindreeper2420:
@aidenkeating ok, so is the idea that we can now use Handlebars elsewhere, as needed? I can see it being useful elsewhere.

PR #177, opened 2018-10-25
Change adoc structure to support copying

finp:


finp:
@dlabaj  could you take another look, I'm finished now

PR #196, opened 2018-11-01
PF4 Branch: Convert PF3 buttons to PF4 buttons

mindreeper2420:
This PR converts the PF3 buttons to PF4 buttons. Each button was converted from a React component to an HTML button, using `className` rather than `<Button` and `bsStyle`.

No additional changes were done as part of this PR.


dlabaj:
Just needs the conflicts resolve.

PR #197, opened 2018-11-01
Remove package-lock

mindreeper2420:
Remove the `package-lock.json` file, in order to only use the `yarn.lock` file. This will eliminate the warning message displayed when running `yarn install`.

Warning message: "warning package-lock.json found. Your project contains lock files generated by tools other than Yarn. It is advised not to mix package managers in order to avoid resolution inconsistencies caused by unsynchronized lock files. To clear this warning, remove package-lock.json."


david-martin:
@mindreeper2420 How about adding package-lock.json to .gitignore, with a comment about why

mindreeper2420:
@david-martin good call - I've updated the `.gitignore` file and added a note in it.

PR #200, opened 2018-11-05
[DO NOT MERGE] Add own walkthroughs

aidenkeating:
This PR is just to show the diffs between master and the adding your own walkthrough changes.

@dlabaj You might be interested in this?


server.js
```diff
@@ -143,4 +205,6 @@ if (process.env.NODE_ENV === 'production') {
   });
 }
 
+loadCustomWalkthroughs('./public/walkthroughs');
```
aidenkeating: This stuff shouldn't be stored in-memory like this

PR #204, opened 2018-11-08
New walkthrough format

aidenkeating:


public/walkthroughs/my-custom-walkthrough/walkthrough.adoc
```diff
@@ -0,0 +1,49 @@
...
+
+The second step needs you to do this other thing.
+
+[type=verification]
+Check that thing.
+
+[type=verificationFail]
+Breathe. Calm down. Try again. Contact your admin.
+
+[type=verificationSuccess]
```
aidenkeating: @david-martin @finp Does thing like like an ok idea to you guys?

If a `verification` has a `verificationFail` or `verificationSuccess`  anywhere between itself and the next `verification` then it will be shown when a user selects `yes/no` in the Radio box?

finp: @aidenkeating  I'm not sure what the point of [type=verificationSuccess] is?
If you succeed, your reward is indicated at bottom of screen with one of required bulbs lighting up (and all the bulbs must be lit for you to click next). But this type is optional, right? 
(apologies for the skeuomorphic design mentions)

david-martin: Is there a need for a success message?
What would that be in different cases, or would it always be the same?

aidenkeating: @finp It is optional, yes. This was mostly added in because it was just the inverse of an error message, so it was pretty simple to add. I can remove it completely if it's something we won't need.

mfrances17: @david-martin @finp @aidenkeating We should probably remove it. I can't really think of a use case for it. I had removed all handling of successverifications from the UI code a while back to simplify the code in task.js.

finp:
@aidenkeating  should we change this line to H3 to match the new taskResource step H3?
https://github.com/integr8ly/tutorial-web-app/blob/7fe0ebae9df0109ef4615c061d5c612995331ebf/src/components/walkthroughResources/walkthroughResources.js#L120

PR #211, opened 2018-11-08
Adding GitHub issue template to the repo

davidffrench:
It can be very useful to have an issue template in GitHub to ensure all the details of the issue are captured when it is created.

Please let me know if you think there should be additional information captured under specifications.


davidffrench:
Thanks for the feedback @mindreeper2420 , I have modified it to include these details now.

PR #242, opened 2018-11-10
WIP: Add existing walkthroughs to feature branch for own walkthroughs

finp:
@aidenkeating just messing around, and looks like there's decent markdown support


public/walkthroughs/1/walkthrough.adoc
```diff
@@ -0,0 +1,318 @@
...
+
+. Click *Send Request*.
+
+. Navigate to the link:http://{spring-boot-url}[Spring Boot app, window="_blank"].
+
+. Check that the entry from the Node.js app is displayed.
+
+
+[type=verification]
+View the activity log:
```
finp: @aidenkeating  this verification doesn't render like I expect, are we supporting multiline verificiations?

aidenkeating: @finp It should support multi-line verifications, but not multi-block verifications. Because we wouldn't know which block is the final block for the verification. Does `+` on the newline resolve the issue?

finp: @aidenkeating gotya, in this case a + does not work, but I can wrap the text in **** and make it a block

public/walkthroughs/markdown-walkthrough/walkthrough.adoc
```diff
@@ -0,0 +1,277 @@
+# A custom walkthrough written in markdown
```
finp: @aidenkeating @david-martin  here's some thinking about markdown support.

david-martin: A few q's.

* Is there anything special the webapp has to do/include to support this (anything to supplement asciidoctor.js)?
* Is it likely *we* will be writing any walkthrough content in markdown?
* What is the benefit of supporting markdown?
* Is it worth trying to support both asciidoc & markdown?

finp: @david-martin 
- asciidoctor supports some markdown by default, no changes required
- no, however many people resist asciidoc as a format, this just eases that conversion
- the story becomes 'this is a variant of markdown', not a new language you must learn
- no, and I'm not sure whether it works as a method of introducing asciidoc, but I feel you are the wrong audience for that test ;)

finp:
closed in favor of https://github.com/integr8ly/tutorial-web-app/pull/262

PR #252, opened 2018-11-14
Updates from Mater

dlabaj:


src/pages/tutorial/task/task.js
```diff
@@ -547,34 +547,31 @@ class TaskPage extends React.Component {
...
-                        <TabPane eventKey={1}>Tab 1 content</TabPane>
-                        <TabPane eventKey={2}>Tab 2 content</TabPane>
-                        <TabPane eventKey={3}>Tab 3 content</TabPane>
+                        <TabPane eventKey={1}>
+                          <Iframe
+                            name="currentApp"
+                            className="integr8ly-iframe"
+                            display="initial"
+                            position="middle"
+                            height="90vh"
```
mindreeper2420: Since there is already a class attached to this element with a property, can the height, display, and position properties be moved there as well? That way everything is contained to the class and the JS doesn't have to be modified for a style update.

PR #255, opened 2018-11-16
Use the next stable version of asciidoctor.js

david-martin:
This version (1.5.7) includes a fix for warnings during webpack builds
See https://github.com/asciidoctor/asciidoctor.js/issues/508


david-martin:
Thanks @mindreeper2420

PR #256, opened 2018-11-16
Walkthrough deps repos

pb82:
## Motivation

Init repository dependencies for custom walkthroughs.

## What

When a custom walkthrough is started the webapp backend will initialize specified gitea repositories. 

## Why

To allow walkthrough authors to specify which gitea repositories should exist for a walkthrough.

## How

When a thread is opened the webapp will send a request to the backend to initialize the dependencies (as specified in `walkthrough.json`). The webapp will verify the users access_token against Openshift and if it's valid create the repos. To to this the webapp has two new env vars: `GITEA_HOST` and `GITEA_TOKEN`. 

## Verification Steps

Needs to be verified together with https://github.com/integr8ly/installation/pull/148


gitea_client.js
```diff
@@ -0,0 +1,165 @@
...
+ * @param username Username
+ * @param repo Repository name
+ * @returns {Promise<any>}
+ */
+exports.createRepoForUser = (username, repo) => {
+  const normalizedUser = normalizeUser(username);
+
+  return new Promise((resolve, reject) => {
+    if (!GITEA_HOST || !GITEA_TOKEN) {
+      return reject(new Error('Gitea not installed'));
```
david-martin: I tried running without setting these env vars, and I got this error in the logs as expected.
However, in the webapp UI it continues past the loading screen to the start of the walkthrough.
It would be better to fail at the loading screen with a useful error.

@aidenkeating Is that something the loading screen supports?

pb82: Makes sense, i will change this to stop at the loading screen.

david-martin:
> Not sure if it's my setup, but I'll debug a little now

Fixed now with https://github.com/integr8ly/tutorial-web-app/pull/256/commits/5aef2bcddab5b4507655fce4b54922bb211aa989

david-martin:
@pb82 this is good to merge after https://github.com/integr8ly/tutorial-web-app/pull/256/files#r234240469

pb82:
@david-martin I've added the changes to display progress when initializing the dependencies.

@mindreeper2420 With this change there are now two distinct progress screen when opening a walkthrough: first you will see a progress bar for initializing the dependencies (repos). Once that reached 100% another progress bar screen will open to provision the service instances.

I reckon it would be better to only have one such screen for repos and service instances. What do you think?

mindreeper2420:
@pb82 I think it'd definitely be better to only one the one screen for the repos and service instances. Is it possible to add any hooks/catches into the process that could allow the progress bar to account for both pieces loading? I think it would be really nice to have the text state what is loading (either a repo or service instance) and change when the next piece starts.

Example: Progress bar starts at 0% and goes to 50% for repo loading (text along the lines of your dependencies are loading) and then 50% - 100% would be the service instances (your services are starting). Thoughts?

I'd also be good with that being a follow-up to this work, so that it doesn't become a blocker.

pb82:
@mindreeper2420 Yes that sounds like a good idea, allowing the text in the progress bar screen to be updated (and trigger that update externally using hooks). I'll try to come up with an implementation but i don't think it's blocking this PR. Since we're not merging the target branch of this straight away i can try to include the work here. Or we can create an extra Trello card for it.

PR #264, opened 2018-11-22
Wt2 new format

pb82:
Walkthrough 2 in the new format

### DIFF
--- a/public/walkthroughs/walkthrough-2/walkthrough.adoc
+++ b/public/walkthroughs/walkthrough-2/walkthrough.adoc
@@ -1,0 +1,138 @@
+= Integrating API-driven applications
+
+:context: integrating-api-driven-applications
+
+Integrate an API that provides flight information, ensuring that the API is protected and rate limited.
+
+APIs are the ideal channel for organizations who want to reach audiences whenever, wherever, and however they choose.
+This enables innovation, mobile and partner development.
+However, it also exposes the hosting infrastructure and organization.
+It is possible to protect and limit access to an API using {api-mgmt-service} as part of your integration.
+
+This walkthrough describes how to create such an integration, using OpenShift, {launcher-service}, {che-service} and {api-mgmt-service}.
+
+++++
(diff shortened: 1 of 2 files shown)


public/walkthroughs/walkthrough-2/walkthrough.adoc
```diff
@@ -0,0 +1,430 @@
...
+<img src="/images/wt2.png" class="img-responsive" alt="integration">
+++++
+
+[time=5]
+
+== Deploying the Fuse Aggregation App
+
+The Fuse Aggregation App provides information about flights, you deploy it using Launcher.
+
+[type=taskResource]
```
pb82: At the moment we have to repeat the task resources for every task. I've added an action item to address this: `Allow walkthrough resources to be specified on the walkthrough overview page (e.g. by providing a [type=walkthroughResource] annotation)`

pb82:
@finp i can still add the `//include::<original doc>[ ]` annotations if you think it makes sense having them

public/walkthroughs/walkthrough-2/walkthrough.adoc
```diff
@@ -0,0 +1,430 @@
...
+. Select the *Configure APIcast* button.
+
+. Select the *Update & test in Staging Environment* button at the bottom again.
+
+. Check that a success message is displayed, and a green line along the left side of the page.
+****
+
+[type=verificationFail]
+Verify that you followed each step in the procedure above.  If you are still having issues, contact your administrator.
+
```
finp: @pb82  think you're missing docs/modules/ROOT/pages/_partials/fuse-aggregation-app-endpoint-activedocs.adoc here

finp: ie in prod, it looks like https://i.imgur.com/wPiP74w.png

pb82: you're right, adding now

PR #266, opened 2018-11-22
Expose custom routes to the end-user walkthroughs

aidenkeating:
## Motivation
Allow ServiceInstance routes that are provisioned to be exposed to the adoc as named attributes.

The names of these attributes is: `route_<resource-metadata-name>_host`.


aidenkeating:
@finp @pb82 Would you mind taking a look? Note that the names of some attributes changed (EnMasse and AMQ credential stuff). You can see the full list of default resources here: https://github.com/integr8ly/tutorial-web-app/pull/266/files#diff-38afe851ac992779c49b7bd0013c3135R74

pb82:
@aidenkeating I tested this with WT2 and it seems to work (https://tutorial-web-app-webapp.apps.pbraun-9f90.openshiftworkshop.com/tutorial/walkthrough-2/task/0).

There is one URL in WT2 that is not resolved by this: `fuse-aggregator-url`. Not sure where this would come from?

src/pages/tutorial/task/task.js
```diff
@@ -234,8 +205,20 @@ class TaskPage extends React.Component {
...
+      verificationClasses = 'alert integr8ly-alert integr8ly-module-column--steps_alert-green';
+      verificationIcon = 'integr8ly-alert-icon far fa-check-circle';
+    }
+    if (isNoChecked) {
+      verificationClasses = 'alert integr8ly-alert integr8ly-module-column--steps_alert-red';
+      verificationIcon = 'integr8ly-alert-icon far fa-times-circle';
+    }
     return (
-      <Alert type="info" className="integr8ly-module-column--steps_alert-blue" key={blockId}>
+      <div className={verificationClasses} key={`verification-${blockId}`}>
```
pb82: Color doesn't change regardless of the selection (see e.g. here: https://tutorial-web-app-webapp.apps.pbraun-9f90.openshiftworkshop.com/tutorial/walkthrough-2/task/0)

aidenkeating: @pb82 I only added this commit recently, do you have it locally? It seems to be working for me

pb82: @aidenkeating Nvm. it's working. The version i deployed didn't have that commit.

pb82:
@aidenkeating I don't see the fuse-aggregator app in any of the walkthrough namespaces. But i don't have any services in `walkthrough.json` and there doesn't seem to be any deployed for WT2 on master either: https://github.com/integr8ly/tutorial-web-app/blob/master/src/services/walkthroughServices.js#L70

aidenkeating:
@pb82 From reading the content of the walkthrough itself. It brings you to Launcher and asks you to provision an app while doing the walkthrough. I think that app that you provision as part of the walkthrough is the Fuse Aggregator app. So the deployment of that service isn't automated.

PR #269, opened 2018-11-23
Global wt resources

pb82:
## Motivation

Add a new annotation: `[type=walkthroughResources]`. Also adds some tests for the custom walkthrough parsing.

## What

Adds a new annotation that allows users to specify 'global' resources that will be displayed on the right hand side of the tutorial overview as well as on the individual tasks. 

## Why

There was no way to specify walkthrough-level resources yet.

## Verification Steps

Check out this branch, start the webapp locally and open the demo walkthrough.

1. Make sure that on the tutorial overview page you can see `A walkthrough resource`
1. Open the first task. You should also see `A walkthrough resource`

## Checklist:

## Progress

- [x] Finished task


package.json
```diff
@@ -80,7 +80,7 @@
     "start:server": "nodemon --watch server.js --exec 'node server.js'",
     "commit:hash": "echo REACT_APP_UI_COMMIT_HASH=$(git rev-list -1 --all) > .env",
     "build": "yarn commit:hash && yarn build:css && yarn build:js && yarn postbuild",
-    "test": "yarn lint && react-scripts test --env=jsdom --coverage",
+    "test": "yarn lint && react-scripts test --updateSnapshot --env=jsdom --coverage",
```
david-martin: Should the snapshots be automatically updated whenever you run the tests?
Doesn't that negate the testing completely if it always updates the snapshots to the latest, or do the tests fail?

Also, I don't think running tests should update src files

aidenkeating:
👀

david-martin:
Verified

image

chrome-extension://nlipoenfbbikpbjkfpfillcgkoblgpmj/edit.html

package.json
david-martin: OK. 
My understanding was these tests fail if the UI for a particular page doesn't match anymore.
This failure lets the developer know something has changed and they have to either:

* change their code as they didn't mean to have the UI be updated
* update the snapshot as they are OK with the highlighted differences

pb82: Ok, that's also a valid viewpoint. I'll change it to a manual step.

src/components/walkthroughResources/walkthroughResources.js
```diff
@@ -63,10 +63,17 @@ class WalkthroughResources extends React.Component {
     return provisioningStatus;
   }
 
+  renderHeading() {
+    if (!this.props.noHeadline) {
+      return <h4 className="integr8ly-helpful-links-heading">Walkthrough Resources</h4>;
```
aidenkeating: Should this still be displayed? In my mind it looks a bit strange to have this heading and then the other headings? @finp @david-martin Any opinions?

PR #273, opened 2018-11-26
WT1A  new format

finp:
## Motivation
convert 1A to custom walkthrough format


public/walkthroughs/1A/walkthrough.json
```diff
@@ -0,0 +1,18 @@
+{
+  "dependencies": {
+    "repos": [{
```
pb82: I think we can remove the repo

pb82:
@finp looks good, only the walkthrough resources are missing

PR #275, opened 2018-11-26
Display thread init error

aidenkeating:
## Motivation
Currently when an error occurs during provisioning we show a generic screen with a message. This change instead displays the error message, so a user has a better idea of how to resolve the issue (Whether to try again or contact an admin).

The main place we were seeing this issue is when trying to init my-custom-walkthrough without having GITEA_HOST or GITEA_TOKEN set. Instead of seeing a generic message, you should now see the error message from the server explaining that Gitea is not configured.

### DIFF
--- a/gitea_client.js
+++ b/gitea_client.js
@@ -124,7 +124,7 @@ exports.createRepoForUser = (username, repo) => {
   const normalizedUser = normalizeUser(username);
   return new Promise((resolve, reject) => {
     if (!GITEA_HOST || !GITEA_TOKEN) {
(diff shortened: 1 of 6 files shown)


aidenkeating:
@pb82 Mind taking a look?

pb82:
@aidenkeating `GITEA_HOST` and `GITEA_TOKEN` are set by the installer. Were they misssing when you ran the installer? But overall it's good to have better error messages 👍

aidenkeating:
@pb82 This was hit when running locally. When using the installer it works as expected.

PR #276, opened 2018-11-26
use single walkthrough resource component

pb82:
This fixes the problem where `No resources available` was displayed twice.

We are now only using one `<WalkthroughResource ... >` component and combine the walkthrough and task level resources in task.js. Also removes the no longer needed `noHeadline` option.


aidenkeating:
👀

PR #277, opened 2018-11-26
Fix wt1 to use correct messaging solution

finp:
## Motivation
incorrect references to amqp in walkthrough 1

## What
add material from old walkthroughs

### DIFF
--- a/public/walkthroughs/1/walkthrough.adoc
+++ b/public/walkthroughs/1/walkthrough.adoc
@@ -90,7 +90,7 @@ Verify that you followed each step in the procedure above.  If you are still hav
 
 // end::creating-api-connector[]
 
-// tag::creating-amqp-connection-in-fuse[]
+// tag::creating-amq-connection-in-fuse[]
 
 // Module included in the following assemblies:
 //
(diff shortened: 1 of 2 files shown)


public/walkthroughs/1/walkthrough.adoc
```diff
@@ -114,41 +114,40 @@ To receive messages from EnMasse, you create a connection in Red Hat Fuse Online
...
+. When prompted with *Select Connector*, select *AMQ Message Broker*.
 
-. Enter the connection URI relating to {enmasse}:
+. When prompted with *AMQ Message Broker Configuration*:
+.. Enter the following into the *Broker URL* field:
 +
 [subs="attributes+"]
 ----  
-amqp://{amq-broker-url}:5672?amqp.saslMechanisms=PLAIN
+tcp://{messaging-broker-url}:61616
```
pb82: @finp @aidenkeating  hmm, according to the list of attributes it should be `amp-url`: https://github.com/integr8ly/tutorial-web-app/blob/add-own-walkthroughs/public/walkthroughs/my-custom-walkthrough/walkthrough.adoc#using-walkthrough-attributes

aidenkeating: @pb82  I think`{amq-url}` is the URL for the AMQ console, this requires the AMQ broker URL. They could probably be named better though

finp: @pb82 using {amq-broker-url}

PR #280, opened 2018-11-27
Hide the example walkthrough by default

david-martin:
Only show it if `SHOW_EXAMPLE_WALKTHROUGH=true` is set.
This is the case for the `start:dev` script

Purposely going with a simple dirname check for a couple reasons:

* avoid unnecessary overhead of reading in extra metadata (i.e. json files) on startup for each walkthrough
* a future feature for labelling walkthroughs, coupled waith a feature for showing/hiding particularly labelled walkthroughs would supercede this


pb82:
👀

david-martin:
Thanks @pb82 .

I'll create a separate PR for the failing build (unrelated)

PR #283, opened 2018-11-27
force copy paste input and buttons to same height

pb82:
The copy and paste input consists of inputs and buttons. Force all of them always to the same height.

Verification steps:

1. Open a few walkthroughs with copy & paste fields. Make sure the buttons and inputs are aligned.
1. Open the example walkthrough (it has a multiline field). This one should also look good (buttons and input aligned).


david-martin:
Verified

Before
image

After
image

PR #284, opened 2018-11-27
Task and steps numbers

finp:
## Motivation
Resolve numbering issue - show numbers so that user can understand status at bottomof page

## Solution
* use `:sectnums:` tag


finp:
@aidenkeating  I think there's an issue with skipping H3 (===), insofar as:
* verifications don't render
* no status hook up to bottom of page

finp:
@pb82 reduced scope to just numbering, applied to all walkthroughs and created https://github.com/integr8ly/tutorial-web-app/issues/287 to address single step tasks

PR #292, opened 2018-11-28
WIP: add instructions to custom walkthrough example

finp:
## Motivation
Custom walkthroughs contain a lot of i8 specific syntax, that needs to be explained
## What
add samples with callouts to explain syntax

## Progress

- [x] draft callout wt
- [x] split into small samples
- [x] overview sample
- [ ] task resources sample
- [ ] walkthrough resources and attributes
- [ ] verfications - fails, and numbering
- [ ] using sectnums

### DIFF
--- a/public/walkthroughs/my-custom-walkthrough/walkthrough.adoc
+++ b/public/walkthroughs/my-custom-walkthrough/walkthrough.adoc
@@ -1,11 +1,47 @@
 = A custom walkthrough
 
-This is the first sentence of the walkthrough's preamble. It will be used as the
-short description for the walkthrough.
+The first sentence is used as the short description for the walkthrough.
 
-These lines will be included in the long description of the walkthrough. Normally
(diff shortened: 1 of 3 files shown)


finp:
@pb82 I added a custom resource, but didn't do it 'properly', could you add and write up how to?

(in new walkthrough, I tried to add 'asciidoc' links to act as walkthrough resource)

pb82:
@finp Was it a rendering problem or did the resource not show up? `[type=walkthroughResource]` will only work in the preamble. It's ignored if you put it inside a task. Could that be the problem? `[type=taskResource]` can be used in tasks. The idea is that a `walkthroughResource` is a global resources that will be added to every task.

pb82:
@finp i'll add this to your document

finp:
closed in favor of https://github.com/integr8ly/example-customisations/pull/9

PR #293, opened 2018-11-28
rely on asciidoc sectnums only for numbering

pb82:
We're using asciidoc `:sectnums:` to automate the numbering. This PR removes the built-in numbering in the tutorial overview to avoid double numbers.

Before:
bildschirmfoto 2018-11-28 um 12 45 16

After:
bildschirmfoto 2018-11-28 um 12 45 37


aidenkeating:
:eyes:

PR #295, opened 2018-11-28
WIP: Add resources for 1 and 1A

finp:
## Motivation
WT2 has resources, but not 1 and 1A

## Progress

- [x] add walkthrough resources for 1

- [ ] add resources for 1A


public/walkthroughs/1/walkthrough.adoc
```diff
@@ -42,6 +42,15 @@ This walkthrough describes how  to create such an integration, using OpenShift,
 // end::master-1[]
 
 
+[type=walkthroughResource,serviceName=fuse]
```
finp: @pb82  how do i add an AMQ resource?
also, how do i know what serviceNames are avail

pb82: @finp this works for all default middleware services which you can find in this list: https://github.com/integr8ly/tutorial-web-app/blob/master/src/common/serviceInstanceHelpers.js#L85

pb82:
👀

public/walkthroughs/1a/walkthrough.adoc
```diff
@@ -40,6 +40,25 @@ This walkthrough describes how  to create such an integration, using OpenShift,
 
 // end::master-1[]
 
+
+[type=walkthroughResource]
```
pb82: Why not use `serviceName=fuse` here?

finp: How about we do a follow on task of https://github.com/integr8ly/tutorial-web-app/issues/307
I think there are a number of items to discuss

PR #298, opened 2018-11-29
fix wt2 attributes

pb82:
Fixes the attributes for WT2

### DIFF
--- a/public/walkthroughs/walkthrough-2/walkthrough.adoc
+++ b/public/walkthroughs/walkthrough-2/walkthrough.adoc
@@ -15,7 +15,7 @@
 :3Scale-ProductVersion: 2.3
 :EnMasse-master-ProductLongName: EnMasse
 :walkthrough: Integrating API-driven applications
-:fuse-flights-aggregator-app-name: fuse-flights-aggregator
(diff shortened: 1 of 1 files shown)


public/walkthroughs/walkthrough-2/walkthrough.adoc
```diff
@@ -15,6 +15,7 @@
 :3Scale-ProductVersion: 2.3
 :EnMasse-master-ProductLongName: EnMasse
 :walkthrough: Integrating API-driven applications
+:fuse-flights-aggregator-app-name: fuse-flights-aggregator
```
pb82: @aidenkeating i logged all the available attributes but there wasn't one for the aggregator app name. Should there be one?

aidenkeating: @pb82 There wouldn't be one for the app name as that's walkthrough specific. There should be one for the route though

pb82: Yes, `route-fuse-flights-aggregator-host` works for the route

aidenkeating:
:eyes:

pb82:
@aidenkeating @mfrances17 thanks for reviewing. I'll give it another speedrun and then merge.

Issue #300, opened 2018-11-29
Bug: GA labels not appearing in Application list


mindreeper2420:
## Description
The GA labels do not appear in the Application list. According to the code, they should be appearing in the same manner that 'preview' and 'community' do.

## Expected Behavior
GA labels should appear in the same manner that the preview and community labels do.

## Actual Behavior
GA labels are not appearing (though they are rendering in the UI).

## Environment

* Operating system: (macOS Mojave
* Browser and version: Chrome Version 70.0.3538.102 (Official Build) (64-bit)
* OpenShift version: N/A
* Yarn version: 1.9.4
* Node version: 10.9.0
* Npm version: 6.2.0
* Project Version/Tag: release/9.9.9 (master)

## Steps to reproduce

1. Start the webapp
2. View the Application list on the dashboard
3. 'GA' should appear next to 3scale, AMQ, and Fuse Online

![screen shot 2018-11-29 at 1 44 20 pm](https://user-images.githubusercontent.com/4032718/49244061-e67b6000-f3dc-11e8-914c-890c9ea8344f.png)

PR #301, opened 2018-11-29
fix(labels): Update labels to show GA and match designs

mindreeper2420:
## Motivation
Fix and close https://github.com/integr8ly/tutorial-web-app/issues/300

## What
The GA labels were not appearing in the UI, although the code is rendering them.

## Why
The CSS had not been updated to handle to GA label.

## How
Add a CSS class to match what the UI is rendering.

## Verification Steps

1. Open the webapp
2. View the dashboard
3. Check the Application list for the GA label

## Checklist:

- [X] Code has been tested locally by PR requester

## Progress

- [x] Finished task

## Additional Notes

This PR also addresses the new [designs](https://redhat.invisionapp.com/d/main#/console/15529764/324779895/preview), where the colors of the labels are simplified to be the same, no matter the label type.
screen shot 2018-11-29 at 1 42 36 pm

### DIFF
--- a/src/styles/application/_labels.scss
+++ b/src/styles/application/_labels.scss
@@ -1,23 +1,13 @@
 .integr8ly-label {
-  &-community {
-    position: absolute;
-    right: var(--pf-global--spacer--lg);
-    padding: var(--pf-global--spacer--xs) var(--pf-global--spacer--sm);
-    border: var(--pf-global--BorderWidth--sm) solid;
-    border-color: $pf-color-black-300;
-    background-color: $pf-color-black-300;
-    color: $pf-color-black-900;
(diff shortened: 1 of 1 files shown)


mfrances17:
This does reflect the original design, and the changes look good, but there was a decision to only show community and preview, and not call out GA (if there was no tag, that meant it was GA) in order to reduce clutter in the panel. Confirmed with Brian - we can probably close this.

Note that there is a related issue that I just filed - issue [304](https://github.com/integr8ly/tutorial-web-app/issues/304), in which GA is showing up in the product names if it appears in the json. I'll add some code that will only display tags if they are preview or community.

PR #308, opened 2018-11-30
Update queue name to work-queue-requests

aidenkeating:
Ensure the broker credentials are provided to the Node app for Walkthrough 1 and 1A.


aidenkeating:
@david-martin Would you mind taking a look? Verification is just ensuring Walkthrough 1 and 1A now work.

PR #314, opened 2018-12-03
Walkthrough locations var

pb82:
## Motivation

Allow walkthrough locations to be passed in via an environment variable (`WALKTHROUGH_LOCATIONS`). If not set defaults to `./public/walkthroughs`. The env var can contain multiple locations separated by semicolon. Also fails if duplicate walkthroughs (by id) are detected.

## Verification steps

1. Run the app in dev mode: it should work as expected
1. Copy your `./public/walkthroughs` directory to some other location (e.g. home directory) and run `WALKTHROUGH_LOCATIONS='<path to walkthroughs>' yarn start:dev`: it should work as expected.
1. Now run `WALKTHROUGH_LOCATIONS='<path to walkthroughs>;./public/walkthroughs' yarn start:dev`: this should fail. Check for a message in the output like `Duplicate walkthrough with id 1 (Build a simple integration that connects a message queue with an API.)`. Note: nodemon will still try to restart the server but the webapp should fail.


aidenkeating:
👀

server.js
```diff
@@ -17,8 +17,12 @@ const configPath = process.env.SERVER_EXTRA_CONFIG_FILE || '/etc/webapp/customSe
 const DEFAULT_CUSTOM_CONFIG_DATA = {
   services: []
 };
+
+const walkthroughLocations = process.env.WALKTHROUGH_LOCATIONS || './public/walkthroughs';
+
 const CONTEXT_PREAMBLE = 'preamble';
 const CONTEXT_PARAGRAPH = 'paragraph';
+const LOCATION_SEPARATOR = ';';
```
aidenkeating: Was there a reason a comma couldn't be used here? (Only asking because I'd be used to comma separated lists)

pb82: A comma would be fine too. I was thinking of bash where a semicolon is used.

PR #321, opened 2018-12-04
allow importing walkthroughs from git repositories

pb82:
## Motivation

Allow importing walkthroughs from git repositories. Git repos can be specified in the `WALKTHROUGH_LOCATIONS` env var. Multiple repos and mixing repos and paths is supported.

The repositories will by default be cloned into a randomized directory under `/tmp`, but the base directory can be overridden using an env var `TMP_DIR`.

## Verification steps

1. run `yarn start:dev`. It's already pointing to included walkthroughs repo (https://github.com/integr8ly/tutorial-web-app-walkthroughs). Everything should work as usual.
1. Check your `/tmp` directory. There should be a randomized subdirectory (uuid) containing the walkthrough.
1. If you run it again, it will create a new directory.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,12 +3,16 @@ FROM bucharestgold/centos7-s2i-nodejs:10.x
 EXPOSE 5001
 
 ENV BUILD_ENV=OCP
+ENV GIT_COMMITTER_NAME=integreatly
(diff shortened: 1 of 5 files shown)


aidenkeating:
:eyes:

aidenkeating:
@pb82 I think we can close this based on #333 ?

PR #331, opened 2018-12-05
Fix: Limit the length of namespace names to prevent Routes that are too long

### LINKED ISSUE #329: URL length still too long in walkthrough 1 results in error in fuse
## Description
A bug we resolved before has been reintroduced. For walkthrough 1 I ended up with a url for the frontend app that looked like 
http://crud-admin-example-com-1-integrating-event-and-api-driven-apps.apps.waterford-59b9.openshiftworkshop.com/v2/api-docs

Which is too long and is in an error state in OpenShift

## Expected Behaviour
A correct url be generated

## Steps to reproduce
Do walkthrough 1 from master 

Suggested Fix

The name of the route doesn't need to be unique across the cluster it only need to be unique to the ns. So you could call it "Frontend" and it would be fine

david-martin:
Some examples of how this will affect namespace names:

`admin-example-com-walkthrough-projects` stays the same

`admin-example-com-1-integrating-event-and-api-driven-apps` becomes
`admin-example-com-1-integrating-eve-HASH`

`evals22-example-com-1-integrating-event-and-api-driven-apps` becomds
`evals22-example-com-1-integrating-e-HASH`

The namespace should still be deterministic.

This change is *NOT* backwards compatible, and will result in a new namespace being created on first login.
This is OK for all newly provisioned clusters.

fixes #329

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,8 @@
     "redux-logger": "3.0.6",
     "redux-promise-middleware": "5.1.1",
     "redux-thunk": "2.3.0",
-    "rimraf": "2.6.2"
+    "rimraf": "2.6.2",
+    "sha.js": "2.4.11"
   },
   "devDependencies": {
     "babel-eslint": "8.2.6",
--- a/src/common/docsHelpers.js
+++ b/src/common/docsHelpers.js
@@ -19,10 +19,10 @@ const getDocsForWalkthrough = (walkthroughId, middlewareServices, walkthroughRes
 
 const getUserAttrs = (walkthroughId, username) => ({
   'openshift-host': window.OPENSHIFT_CONFIG.masterUri,
-  'project-namespace': buildValidProjectNamespaceName(username, 'walkthrough-projects'),
+  'project-namespace': buildValidProjectNamespaceName(username, 'shared'),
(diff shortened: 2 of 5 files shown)


david-martin:
@aidenkeating interested in your thoughts on the `<email>-walkthrough-projects` project that gets created for each user.

With this change, the project name will potentially be truncated too.
e.g. `evals222-example-com-walkthrough-projects` becomes `evals222-example-com-walkthrough-pro-HASH`

An edge case, but wondering what you think.
It wouldn't be a bug or blocker either way

aidenkeating:
@david-martin I think it depends on whether there's a realistic chance of the users email ever causing the projects namespace to go above the max length? If it's not then I'd prefer to keep the projects namespace the way it was because it distinguishes it from the others, but that's more personal preference than anything.

Also, this means there cannot be a walkthrough named `walkthrough-projects` (which is very unlikely anyway).

david-martin:
@aidenkeating I've changed the suffix to `-shared`. Would you mind re-reviewing?

david-martin:
shared project changes
image

david-martin:
WT1A with fix

image

PR #332, opened 2018-12-06
Remove unused walkthrough specific code

david-martin:
Remove unused walkthrough specific code. This code has been replaced by generic code that reads in metadata from walkthrough.json per walkthrough


david-martin:
@aidenkeating could you explain the difference between this block https://github.com/integr8ly/tutorial-web-app/blob/master/src/services/middlewareServices.js#L28-L36

and this block
https://github.com/integr8ly/tutorial-web-app/blob/master/src/components/provisioning/provisioning.js#L10-L17

They look to be doing something similar.

aidenkeating:
@david-martin I can't remember the original difference between them, but I think one of them can be removed now. They do the same thing now, list the services that should be provisioned on login.

PR #335, opened 2018-12-07
Reduce namespace length down to 20 characters

aidenkeating:
Currently, we are seeing issues with route lengths going above the
maximum limit. This is partially because namespace lengths are too
long. This change reduces the namespace length down to 20 characters
with the format {trimmedUsername}-{trimmedSuffix}-{hash} where:

- trimmedUsername is 10 characters or less
- hash is 4 characters
- trimmedSuffix makes up the remaining characters


david-martin:
Verified. Much better experience

image

PR #336, opened 2018-12-07
Add an endpoint for re-syncing walkthroughs

aidenkeating:
Adds a `POST /sync-walkthroughs` to re-clone the repositories.


pb82:
@aidenkeating When i call the endpoint i get the expected JSON response. But when i refresh the tutorial web app i still see the old content, so it looks like that's not updated.

pb82:
@aidenkeating Ah nvm. that was because i was testing on a cluster which is apparently production mode

PR #342, opened 2018-12-12
Intly 184

pb82:
Changes the Syntax for Walkthrough and Task resources. Previously we required the use of third level headings for resources. This conflicts with the Asciidoc standard because a walkthrough resource placed in the Preamble (a h1) can't directly contain a h3 element.

The new Syntax uses delimited blocks:

```asciidoc
[type=walkthroughResource,serviceName=openshift]
.Openshift
****
* link:{openshift-host}/console[Console]
****
```

Verification steps:

1. Create a Walkthrough that uses the new resource format
1. Import that walkthrough locally using `WALKTHROUGH_LOCATIONS`
1. Make sure the resources are rendered correctly

Before this can be merged we need to update all the walkthroughs in https://github.com/integr8ly/tutorial-web-app-walkthroughs to the new format.


finp:
@pb82  will follow up to change all walkthroughs to new format

PR #351, opened 2019-01-02
INTLY-107 Allow single step tasks without a title

aidenkeating:
## Motivation
Currently, when creating a Walkthrough if a writer would like to
have a task that contains a single step they must still define a
h3 block and give the block a title. Otherwise features like the
verification block cannot be used.

In some cases, when creating this type of single step task, the
title of the task is the title of the step also.

## What
This change allows tasks to contain verification blocks, instead
of them only being allowed in steps. Meaning that a writer can
create a single step task without having to specify a h3 block.

Verification blocks will still work as expected in the web app.

## How
The `WalkthroughTask#fromAdoc` function now also allows for verifications to be included at its top level, instead of needing to be nested inside steps.

## Verification Steps

* Add a task that looks something like the following to an existing walkthrough, and ensure it renders correctly and that the `Next` button only goes blue when all verifications are complete.

```
== Example task

An example task with no steps, but it includes verification blocks.

[type=verification]
This should be the first verification

[type=verificationFail]
This should be the first verification fail message

[type=verification]
This should be a second verification

[type=verificationFail]
This should be a second verification fail message
```
* Ensure the rest of the walkthrough verifications show and work as expected

### DIFF
--- a/src/common/walkthroughHelpers.js
+++ b/src/common/walkthroughHelpers.js
@@ -267,6 +267,10 @@ class WalkthroughTask {
     return this._steps.filter(s => !(s instanceof WalkthroughResourceStep));
   }
 
+  get blocks() {
+    return this._steps;
+  }
+
   get resources() {
     return this._resources;
   }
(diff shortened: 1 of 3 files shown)


aidenkeating:

Here's a link to the Jira issue: https://issues.jboss.org/browse/INTLY-107

PR #352, opened 2019-01-02
INTLY-380 clicking logo does hard refresh

aidenkeating:
## Why
Currently, when the "Red Hat Solution Explorer" logo in the top
left-hand corner is clicked we set window.href explicitly, this
can result in a full refresh of the page, which then results in
the loading screen for the web app revealing itself momentarily.

## How
We do not need to perform this kind of full refresh normally.
Instead we can use the react router to change the state of the app
so that it's on the landing page, this results in a much faster
transition without a loading screen.

## Verification Steps
* On the landing page, select "Red Hat Solution Explorer"
* The page should appear again immediately
* Do the same from a Walkthrough overview page to ensure the loading
screen doesn't appear.

### DIFF
--- a/src/components/masthead/__tests__/__snapshots__/masthead.test.js.snap
+++ b/src/components/masthead/__tests__/__snapshots__/masthead.test.js.snap
@@ -2,6 +2,11 @@
 
 exports[`Masthead Component should render a basic component: basic 1`] = `
 <Masthead
+  history={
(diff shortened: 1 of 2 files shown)


aidenkeating:

Here is a link to the Jira issue: https://issues.jboss.org/browse/INTLY-380

PR #354, opened 2019-01-04
INTLY-416 Fix duplicate task resources in single step tasks

aidenkeating:
JIRA: https://issues.jboss.org/browse/INTLY-416

## Motivation
Currently, when defining a single step task that includes a task
resource block it will be included on both the right hand panel
and in the main content.

This is because the task resource blocks are not filtered out in
single step tasks.

## How
This is resolved by filtering out the task resource blocks when
defining the main body blocks for the task.

## Verification Steps
* Have a single step task with a taskResource block inside it
* Ensure the taskResource block is rendered on the right panel and
not in the main body
* Ensure the normal, multi-step, steps work as expected

## Checklist:


aidenkeating:
@finp Mind taking a look?

PR #358, opened 2019-01-09
INTLY-418 Use new loading screen

aidenkeating:
## Motivation
The current loading screen being used does not display enough
information about the services being provisioned, the percentage
usually stalls at a percentage as one service takes much longer to
provision than others. This leads the end-users of the web app to
think that something has gone wrong.

## What
The new loading screen has a per-service status view which is a
better indicator of what's going on in the background.

## Verification Steps
* Login to the web app as a user with no Shared namespace
* Ensure the loading screen works as expected


aidenkeating:
@mindreeper2420 @mfrances17 I seem to still have some missing styling, not sure where it should be coming from though. Notice how the width of the status bars isn't full screen and how it doesn't stick to the bottom of the screen.

screen shot 2019-01-09 at 09 02 32

david-martin:
Will take a look locally to see whats up with the bottom padding/margin

src/components/provisioning/provisioning.js
```diff
@@ -98,6 +99,125 @@ function buildProvisioningScreen(WrappedComponent) {
...
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+        </div>
+      );
+    }
+
+    static newLoadingScreen(services) {
```
aidenkeating: This needs to be renamed and the old functions removed

david-martin:
> how it doesn't stick to the bottom of the screen.

Pushed fix in https://github.com/integr8ly/tutorial-web-app/pull/358/commits/02c4f336aed047aa7bb629e57d93617c07f9e6bd

mindreeper2420:
@aidenkeating can you change the text for 'Ready to use' and 'Error' to white, removing the green/red? We noticed that they don't have enough contrast to be truly readable. The icons and indicator bars can stay as their respective colors. Thanks!

aidenkeating:
@david-martin @mindreeper2420 @mfrances17 Mind taking another look? I've included a screenshot below of what it looks like.

@mindreeper2420 Should the red be kept for the error state on the blue, or should it be black or something?

screen shot 2019-01-09 at 15 01 29

aidenkeating:
@david-martin I've resolved the cursor issue you saw and resolved the failed tests, mind taking another look?

@mindreeper2420 @mfrances17 @david-martin I'll be offline for a couple of hours, if there are any changes needed I'll address them this evening

PR #360, opened 2019-01-09
[WIP] INTLY-473 Ensure progress is never stored greater than 100%, and will never ren…

david-martin:
…der greater than 100%

## Progress

- [x] Avoid progress being shown as greater than 100%
- [ ] Invalidate state in local storage if number of steps has changed in a walkthrough


mfrances17:
This PR is no longer needed... this issue was just recently fixed in a later version of the code referencing the same issue INTLY-473.

PR #362, opened 2019-01-09
Remove hardcoded text & unset var

david-martin:
Before (hardcoded walkthrough text):
image

After:
image


tremes:
Maybe just remove the `threadName` variable to make CI happy.

PR #371, opened 2019-01-21
INTLY-383 If preamble doesn't have lines ignore shortDescription

aidenkeating:
JIRA: https://issues.jboss.org/browse/INTLY-383

Currently, in the server-side parsing of the walkthrough files, if
the walkthrough has a preamble that does not start with a paragraph
containing at least one line then the parsing will fail. This is
due to the parser expecting that a walkthrough preamble being a
paragraph.

This change performs a check on the preamble to ensure that it
checks before trying to retrieve the shortDescription that the
preamble starts with a paragraph that contains at least one line.

Verification:
- Put a bulleted list as the preamble of a paragraph (first content
after the title)
- Restart the server and ensure it does not error
- Go through the walkthrough, ensure it works and the overview
renders correctly

### DIFF
--- a/server.js
+++ b/server.js
@@ -371,7 +371,11 @@ function getWalkthroughInfoFromAdoc(id, dirName, doc) {
   // ````
   // So it's better to just tell the user to put a blank line between the title and short description
   let shortDescription = '';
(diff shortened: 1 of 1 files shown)


aidenkeating:
@finp Would you mind giving this a try and seeing whether it solves the issue you were seeing?

finp:
:eyes:

PR #374, opened 2019-01-22
INTLY-379 Stop loading screen between tasks in walkthrough

aidenkeating:
Currently when a user selects to go to the next task in a
walkthrough a loading screen will appear momentarily. This is due
to the request for the asciidoc contents and manifest of the
walkthrough being requested each time the task component mounts.
This request clears the current Redux store, causing the loading
page to appear until the store is populated again.

However, when moving between tasks in the same walkthrough this
does not need to be done as long as the asciidoc content and the
manifest have already been retrieved and stored in Redux.

This change performs a check before trying to kick off the
requests to retrieve this information. The check performed when
the task component mounts is:
- Is this task for the same walkthrough as the last task
- Is the asciidoc content already available in the Redux store
- Is the manifest content already available in the Redux store

If these are true, the loading screen does not need to be shown.

Verification:
- Go into any walkthrough with more than one task
- Click through the tasks of the walkthrough, ensure the loading
screen is not shown
- Click on the 'home' icon in the breadcrumbs to go to the landing
page
- Click on any other walkthrough and ensure the loading screen is
shown (it's a new walkthrough, we need to load again)
- Perform a hard refresh on a task, ensure that a loading screen
is shown (we have no knowledge of the last walkthrough)


aidenkeating:

PR #375, opened 2019-01-22
Add release steps for webapp & walkthroughs

david-martin:


david-martin:
@mikenairn would you mind reviewing the steps we used for releasing?

PR #378, opened 2019-01-23
Fix changing line height on provisioning screen

matskiv:
## Motivation
When service provisioning finishes, the icon on the left is changed, row height is changed and this causes vertical expansion of the rows - looks like they are jumping :)

## What
Fix style for integr8ly-status-ready css class.

## Why
To improve UX.

## How
a) change font-size, with 1.6em "ready" icon had height of 25.6px. Loading image has height of 24px
b) change display to block to avoid unnecessary empty space in the parent element

## Verification Steps
Open provisioning screen for a new user.
Height of the rows should not change when they transition from "Provisioning" to "Ready to use" 

## Checklist:

## Progress

- [x] Finished task

## Additional Notes
Before:
before
After:
after

### DIFF
--- a/src/styles/application/_provisioning.scss
+++ b/src/styles/application/_provisioning.scss
@@ -164,6 +164,8 @@
 
 .integr8ly-status-ready {
   color: $pf-color-light-green-400 !important;
+  font-size: 1.5em !important;
+  display: block;
 }
 
 .integr8ly-status-error,


mindreeper2420:
@matskiv LGTM - feel free to merge.

matskiv:
@mindreeper2420 Thanks for the review!

Issue #392, opened 2019-02-05
Allow the use of customized context dir for the walkthroughs


hguerrero:
## Description

Currently you are forced to use "walkthoughs" as the context dir for your walkthroughs (WTs) repos, when in our workshops we have only one repo with a different layout. So having a customizable CONTEX_DIR var makes sense.
## Expected Behavior

Use the CONTEXT_DIR env var to specify where to look for the WTs.
## Actual Behavior

WTs must be in "walkthroughs" dir.
## Environment
master
* Operating system: (E.g RHEL 7.6 )
* Browser and version: (E.g Google Chrome Version 70.0.3538.102 (Official Build) (64-bit))
* OpenShift version:

* Yarn version:

* Node version:

* Npm version:

* Project Version/Tag: (E.g release-1.0.1)

evanshortiss:
This seems reasonable, but I think the PR in #393 needs some changes to work. Namely, the `WALKTHROUGH_CONTEXT_DIR` being used is applied to *all* walkthroughs.

Instead it should be mapped on a 1:1 basis so each walkthrough can have a `WALKTHROUGH_CONTEXT_DIR`. Maybe instead of using an environment variable we include an `integreatly.json` (or add an `integreatly` section into a `package.json`) in the walkthrough repository like so:

```js
{
  // We could use this as a topic/section name to create unique walkthrough groups
  "title": "My Walkthroughs",
  "description": "A set of walkthroughs that will teach you to use X",
  "walkthroughsFolder": "./custom/walkthrough-path"
}
```

Thoughts @aidenkeating @david-martin @hguerrero?

aidenkeating:
To me, a context-dir-per-repo mapping would be the solution if we want to move away from the hard `walkthroughs/` convention. Otherwise having one repo of walkthroughs with a different structure would not allow others to be imported.

A hacky way would be to allow it to be specified in a query string of the URL or something, then just parse it out on the server. That's not very pretty though.

A concern I'd have about something like `integreatly.json` is where does it go in the repo? I'd want to be able to specify/override it, then it leads to the same issue, that's probably just me though.

I'd be happy with any context-dir-per-repo solution, or leaving it as a strict convention as I'm not sure how much time we're allocating to the webapp at the moment.

@david-martin Probably has a better insight on this though.

evanshortiss:
[...]

PR #393, opened 2019-02-05
make context_dir an external env var so you can use something like /d…

hguerrero:
…ocs/labs instead of just 'walkthroughs'

## Motivation

#392 
## What

added CONTEXT_DIR env var with default to walkthroughs
## Why

needed for DIL workshops
## How

adding env var, then when joining the path use the value from var instead of hardcoded text.
## Checklist:

- [ ] Code has been tested locally by PR requester

## Progress

- [x] Finished task
- [ ] TODO


evanshortiss:
Handled by PR here - https://github.com/integr8ly/tutorial-web-app/pull/402

Issue #396, opened 2019-02-06
Image Responsiveness


evanshortiss:
## Description
Currently images need `integr8ly-img-responsive` to be added in the walkthroughs to prevent them from overflowing the content pane and cusing layout strangeness (see attached)

## Expected Behavior
After chatting with some others, we concluded expect images to automatically size to the content pane.

## Actual Behavior
Images overflow if the `role="integr8ly-img-responsive"` is not added in the walkthrough `adoc` (see attached)

## Environment

* Operating system: n/a
* Browser and version: n/a
* OpenShift version: n/a
* Yarn version: n/a
* Node version: n/a
* Npm version: n/a
* Project Version/Tag: Current HEAD - ff38fa88518bc10eb05fc44e8bbe9313e0a237e8

## Steps to reproduce
Insert a large image, e.g `image::https://large.images.com/image.png[integration, role="integr8ly-img-responsive"]` into a walkthrough asciidoc and view it in the Solution Explorer.

evanshortiss:
<img width="1552" alt="screenshot 2019-02-05 at 3 55 48 pm" src="https://user-images.githubusercontent.com/1303687/52312667-36881d00-2960-11e9-880a-bb6e1d172dc2.png">

evanshortiss:
With the PR applied, you can see in this screenshot that the `img-responsive` class is not applied, yet we can see all of this beautiful Norwegian landscape.

<img width="1552" alt="screenshot 2019-02-05 at 5 10 04 pm" src="https://user-images.githubusercontent.com/1303687/52314861-51ab5a80-2969-11e9-9ccc-5db59026e727.png">

PR #397, opened 2019-02-06
feat: make images responsive by default

evanshortiss:
## Motivation
Addresses #396 

## What
* Added a new class for the task overview page.
* Updated styles to make images width to max out at 100% of the container width instead of overflow.
* Changed `img-responsive` to `img-not-responsive` so authors can make images overflow if that's somehow necessary

## Why
See #396 

## How
Answered above in "What"

## Verification Steps
1. Update the core walkthroughs repo to remove all `role="integr8ly-img-responsive"` from images
2. Run the webapp against updated core walkthroughs
3. Verify images are constrained to walkthrough content pane width
4. Add `role="integr8ly-img-not-responsive"` to a wide image file in a walktrough
5. Load the walkthrough and shrink the browser window. The image should be obscured by the right hand walkthrough resources panel

## Checklist:

- [ ] Core walkthroughs have been updated (optional, this can be done later since the walkthroughs as they are will still look fine)

## Progress

- [x] Finished task

### DIFF
--- a/src/pages/tutorial/__tests__/__snapshots__/tutorial.test.js.snap
+++ b/src/pages/tutorial/__tests__/__snapshots__/tutorial.test.js.snap
@@ -57,6 +57,7 @@ exports[`TutorialPage Component should render the TutorialPage component fulfill
           </Button>
         </div>
         <div
+          className="integr8ly-task-overview"
           dangerouslySetInnerHTML={
             Object {
               "__html": "<div class=\\"literalblock\\">
--- a/src/pages/tutorial/tutorial.js
+++ b/src/pages/tutorial/tutorial.js
@@ -92,7 +92,7 @@ class TutorialPage extends React.Component {
                   </Button>
                 </div>
                 {this.renderPrereqs(thread)}
-                <div dangerouslySetInnerHTML={{ __html: parsedThread.preamble }} />
+                <div className="integr8ly-task-overview" dangerouslySetInnerHTML={{ __html: parsedThread.preamble }} />
                 {/* <AsciiDocTemplate
                   adoc={thread}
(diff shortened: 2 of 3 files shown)


mindreeper2420:
@evanshortiss thanks for the PR!

In regards to your comment "Changed `img-responsive` to `img-not-responsive` so authors can make images overflow if that's somehow necessary",  based off of the designs, the images should never overflow the content area, always responding to the container. Even though this does constrict the user, it was done so that users would not run into any issues with the Resources area, especially on smaller devices (we test on browser sizes of 1366 x 768 as a default). The change you have proposed for targeting the the image and setting the `max-width` looks good to me.

In the branch **[PF4](https://github.com/integr8ly/tutorial-web-app/tree/pf4)** (which will become the new base for the Master branch in the near future), a lot of the content areas and constraints have been updated. If interested, feel free to take a look at the branch.

evanshortiss:
> based off of the designs, the images should never overflow the content area

@mindreeper2420, great, glad to know we're not the only ones thinking so!

How close is the `PF4` branch with that change to landing? Right now people authoring walkthroughs need to use the `img-responsive` role/class and some folks are migrating a large amount of content. If we could avoid having to add that class/role it would be great.

mindreeper2420:
The goal is to have the `PF4` branch merged into `Master` by the end of month. I don't know when that will make it into a release though.

We can merge this one if needed sooner, and then just recheck once the `PF4` branch is going to be merged in.

evanshortiss:
> We can merge this one if needed sooner

@mindreeper2420 that would be ideal. The folks doing DIL and other workshops would appreciate it 🙏

PR #402, opened 2019-02-11
INTLY-948 feat: add ability to specify custom path to walkthroughs in git url

evanshortiss:
## Motivation
See #392. Essentially some existing content being ported is in a particular format and can often contain multiple "tracks" of work in a single repo. To avoid splitting these repos up we should support a walkthroughs directory override.

## What
Added a parsing step for Git URLs passed via `WALKTHROUGH_LOCATIONS` environment variable. This will extract querystring params from the Git URL.

## Why
See motivation.

## How
By allowing users to override the default `walkthroughs/` path.

## Verification Steps
1. Using this branch start the application like so:

```
WALKTHROUGH_LOCATIONS="https://github.com/RedHatWorkshops/dayinthelife-integration?walkthroughsFolder=docs/labs/citizen-integrator-track&walkthroughsFolder=docs/labs/developer-track/#V2,/Users/eshortis/workspaces/work/integreatly/tutorial-web-app-walkthroughs/walkthroughs" yarn start:dev
```

2. The walkthroughs from `docs/labs/citizen-integrator-track/` for the repo `https://github.com/RedHatWorkshops/dayinthelife-integration#V2` should load in the UI despite not being in the typical `walkthroughs/` folder. The same should occur for `docs/labs/developer-track/` from the same repo. Also, note the URL when you load these walkthroughs - it contains the parent folder name, and the specific lab name to prevent conflicts since both contain `labNN` naming conventions.

## Checklist:

- [x] Changes have been successfully verified by another team member

## Progress

- [x] Finished task

### DIFF
--- a/server.js
+++ b/server.js
@@ -9,12 +9,10 @@ const giteaClient = require('./gitea_client');
 const gitClient = require('./git_client');
 const bodyParser = require('body-parser');
 const uuid = require('uuid');
-<<<<<<< HEAD
 const promMid = require('express-prometheus-middleware');
 const Prometheus = require('prom-client');
-=======
 const querystring = require('querystring')
->>>>>>> feat: add ability to specify custom path to walkthroughs in git url
+const flattenDeep = require('lodash.flattendeep')
 
 const app = express();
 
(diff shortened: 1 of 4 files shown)


server.js
```diff
@@ -164,24 +166,61 @@ function resolveWalkthroughLocations(locations) {
...
+        const walkthroughParams = querystring.parse(url.parse(location).query)
+
+        console.log(`Importing walkthrough from git ${cloneUrl}`);
         return gitClient
-          .cloneRepo(location, clonePath)
+          .cloneRepo(cloneUrl, clonePath)
           .then(cloned => {
-            const locationResult = Object.assign({}, locationResultTemplate, { local: path.join(cloned, 'walkthroughs') });
-            return resolve(locationResult);
+            if (walkthroughParams.walkthroughsFolder && Array.isArray(walkthroughParams.walkthroughsFolder)) {
```
evanshortiss: There can be multiple `walkthroughsFolder` params in the querystring, so need this array check.

server.js
```diff
@@ -164,24 +166,61 @@ function resolveWalkthroughLocations(locations) {
...
+            }
+            // return resolve(locationResult);
           })
           .catch(reject);
       }
       return reject(new Error(`${location} is neither a path nor a git repo`));
     });
   });
+  
+  return Promise.all(mappedLocations).then(flattenDeep);
```
evanshortiss: We perform the flatten on the results due to the above Array of `walkthroughsFolder`

server.js
```diff
@@ -164,24 +166,61 @@ function resolveWalkthroughLocations(locations) {
...
-        const locationResult = Object.assign({}, locationResultTemplate, { local: location });
+        const locationResult = Object.assign({ parentId: 'local' }, locationResultTemplate, { local: location });
         return resolve(locationResult);
       } else if (isGitRepo(location)) {
-        console.log(`Importing walkthrough from git ${location}`);
         const clonePath = path.join(TMP_DIR, tmpDirPrefix);
+
+        // Need to parse out query params for walkthroughs, e.g custom directory
+        const cloneUrl = generateCloneUrlFromLocation(location)
+        const repoName = retWalkthroughRepoNameFromLocation(location)
```
evanshortiss: Including this will prevent clashes in IDs across walkthroughs, even if they use the same ID, e.g `walkthrough-1` or `lab01`.

pb82:
@evanshortiss I'm getting `TypeError: window.OPENSHIFT_CONFIG is undefined` when using the provided command. But it seems to work when i just run `yarn start:dev`. Looking into that.

evanshortiss:
@pb82 I get that at first too. I think it's because it takes a while to clone? If you wait and refresh it seems to resolve once the cloning of repos completes.

pb82:
Verified locally. I'll give it a try on a cluster and then approve.

server.js
finp: is that just for one repo, or all repos specified in walkthrough_locations?

finp: @evanshortiss ping re. this question

tiffanynolan:
UXD is ok to have this merged. Thanks for the heads up about the changes, so we can keep this in mind as we work on groupings and the walkthroughs order for the dashboard. 

@mfrances17 FYI

server.js
evanshortiss: @finp do you mean can two walkthroughs in two repos use the same ID? If so, then yes. For example

```
-- REPO_1
  -- walkthroughs
    -- lab01

-- REPO_2
  -- walkthroughs
    -- lab01
```

Would work since the generated ID is `REPO_1-lab01` and `REPO_2-lab01`

evanshortiss: Oh. I see what you mean. Yes, if it's a single repo with multiple folder then it also works, e.g

```
-- REPO
  |-- set-a
     |-- lab01
  |-- set-b
     |-- lab01
```

Would generate an ID `REPO-set-a-lab01` or similar.

finp: thanks for the clarification, and yes, this does address the original Jira ;)

evanshortiss:
@finp @tiffanynolan @pb82 are we able to merge this for now while work on the new proposal happens?

finp:
@evanshortiss That works for me, whether this method is officially documented is still an open question, but I'm not blocking the code change.

PR #405, opened 2019-02-13
remove amq from webapp

pb82:
AMQ is now AMQ Online but the Webapp was still trying to provision AMQ. This PR removes AMQ and updates the mock config. The Webapp should work in dev mode `yarn start:dev` as well as on a cluster.


server.js
```diff
@@ -287,7 +287,7 @@ function getMockConfigData() {
         },
         {
           spec: {
-            clusterServiceClassExternalName: 'amq-broker-72-persistence'
+            clusterServiceClassExternalName: 'amq-online-standard'
```
sedroche: I was expecting all removals. Why do we point at a different AMQ?

sedroche:
Verified with https://github.com/integr8ly/installation/pull/349

server.js
pb82: AMQ has been replaced with AMQ Online (which as far as i understand comes with the cluster by default).

sedroche: :+1:

aidenkeating: Looks like there's two `amq-online-standard`s listed here now?

pb82: right, i'll remove that. It's the mock config, so can be verified with `yarn start:dev`

PR #412, opened 2019-02-19
[WIP] INTLY-600 updates to integreatly webapp to include prometheus metrics

Nanyte25:
## Motivation

https://issues.jboss.org/browse/INTLY-600
## What
 Update to webapp to expose metrics endpoint to support Prometheus monitoring and metrics.

## Why
This is a needed and supported feature. 

## How

By including the node lib prom-client and defining the metrics end-point.

## Checklist:

## Progress

- [ ] Finished task
- [ ] TODO

### DIFF
--- a/server.js
+++ b/server.js
@@ -3,16 +3,29 @@ const path = require('path');
 const url = require('url');
 const fs = require('fs');
 const asciidoctor = require('asciidoctor.js');
-const adoc = asciidoctor();
 const Mustache = require('mustache');
 const { fetchOpenshiftUser } = require('./server_middleware');
 const giteaClient = require('./gitea_client');
 const gitClient = require('./git_client');
 const bodyParser = require('body-parser');
 const uuid = require('uuid');
+const promMid = require('express-prometheus-middleware');
+const Prometheus = require('prom-client');
 
 const app = express();
+
+const adoc = asciidoctor();
+
 app.use(bodyParser.json());
+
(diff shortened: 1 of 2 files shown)


metrics.js
```diff
@@ -0,0 +1,28 @@
...
+
+// Create a new metric counter
+const counter = new client.Counter({
+	  name: 'metric_name',
+        help: 'metric_help'
+});
+require('prom-client').collectDefaultMetrics({ register });
+
+// New recordcount 
+function recordcount() {
```
david-martin: I'd say you can remove this counter and just have the default metrics for now.

server.js
```diff
@@ -76,6 +77,12 @@ app.get('/config.js', (req, res) => {
   }
 });
 
+// new metrics endpoint for Walkthrough applciation.
```
david-martin: Minor thing, but maybe a more timeless comment like `// prometheus metrics endpoint` would be better here

metrics.js
Nanyte25: @david-martin happy to remove the counter it was just a test. As the other npm module above and will test it on a cluster on see. Happy to go with the alternative after the test.

server.js
damienomurchu:
```suggestion
// new metrics endpoint for Walkthrough application.
```
(if leaving the comment)

PR #414, opened 2019-02-19
fix(masthead): remove masthead per usability results

mindreeper2420:
## Motivation
Remove the Masthead from the dashboard, as it rated poorly in usability tests due to the dark background.

## What
Remove Masthead from the dashboard.

## Why
Usability results.

## How
Delete the code.

## Verification Steps

1. Navigate to the dashboard
2. The masthead should no longer appear

## Checklist:

- [X] Code has been tested locally by PR requester

## Progress

- [x] Finished task

## Additional Notes

screen shot 2019-02-19 at 5 14 48 pm


mindreeper2420:
@tiffanynolan This PR removes the masthead from the dashboard - I've added a screenshot of the new dashboard in the description above.

src/pages/landing/landingPage.js
```diff
@@ -17,22 +17,10 @@ class LandingPage extends React.Component {
   render() {
     const { walkthroughServices, middlewareServices, user } = this.props;
 
-    const LandingPageMastHead = () => (
```
mfrances17: Do you think we should use this opportunity to remove landingPageMastHead.js as well as part of this change? I think it was obsolete even before this change.

mindreeper2420: Good call - I've removed that file.

PR #420, opened 2019-02-25
add walkthrough-locations attribute

finp:
To support 'Publishing' walkthrough, user needs to understand the current configuration before changing it. This PR allows us to include the current config in walkthrough

### DIFF
--- a/src/common/docsHelpers.js
+++ b/src/common/docsHelpers.js
@@ -25,7 +25,8 @@ const getUserAttrs = (walkthroughId, username) => ({
     walkthroughId || buildValidProjectNamespaceName(username, 'shared')
   ),
   'user-username': username,
-  'user-sanitized-username': cleanUsername(username)
+  'user-sanitized-username': cleanUsername(username),
+  'walkthrough-locations': process.env.WALKTHROUGH_LOCATIONS
 });
 
 const getWalkthroughSpecificAttrs = (walkthroughId, walkthroughResources) =>


src/common/docsHelpers.js
```diff
@@ -25,7 +25,8 @@ const getUserAttrs = (walkthroughId, username) => ({
     walkthroughId || buildValidProjectNamespaceName(username, 'shared')
   ),
   'user-username': username,
-  'user-sanitized-username': cleanUsername(username)
+  'user-sanitized-username': cleanUsername(username),
+  'walkthrough-locations': process.env.WALKTHROUGH_LOCATIONS
```
finp: @aidenkeating  this isn't working for me, any ideas?

aidenkeating: @finp I believe `docsHelpers.js` is only used from the client-side? The browser wouldn't have access to the env var by default. We could make a change to pass it from the server to the client if we need it.

finp: @pb82 seems like we need a new plan? But you mentioned that there's a plan to add default value to tutorial web app operator, right? That would also solve my issue in docs, if there's a jira?

pb82: @finp ah right, sorry. I forgot that that code is only ever run on the client. Looking for that JIRA now.

finp:
closing as useless :wink:

Issue #410, opened 2019-02-18
servicesToProvision is not working as expected


aidenkeating:
## Description
Currently, in the webapp, we can provide a JSON file that overrides the services that should be provisioned by the webapp when a user logs in for the first time. If a user doesn't provide this list, then the following services are provisioned:

* AMQ Online (`amq-online-standard`)
* Fuse (`fuse`)
* CodeReady (`codeready`)
* Launcher (`launcher`)
* Apicurito (`apicurio`)

The ID's in brackets are what's provided in the JSON file. For example, a user could have a file:

/etc/webapp/customServerConfig.json
```
{
    "servicesToProvision": ["fuse"]
}
```

The issue we are seeing is that when the user overrides `servicesToProvision` we see only the services we specify in the loading screen. However, when we hit the webapp landing page, we can still see all the services listed on the right-hand side.

## Expected Behavior
Only the services specified in `servicesToProvision` should be shown on the right-hand side.

## Actual Behavior
All services are shown on the right-hand side. I'd guess ones that aren't specified in `servicesToProvision` also have broken links, I haven't confirmed that.

## Steps to reproduce

* Run the webapp pointed at a cluster, specify a `SERVER_EXTRA_CONFIG_FILE` env to point to a `customServerConfig.json` like the one above.
* Login to the webapp as any user
* In the loading screen you should only see the services specified listed
* On the landing page all services will be listed, even though only a subset are specified

aidenkeating:
@tonyxrmdavdison

aidenkeating:
@tonyxrmdavidson

PR #424, opened 2019-02-26
[Tony Davidson]  Fixed reference to customServerConfig.json in server…

tonyxrmdavidson:
….js. Updated READme to give more detail to the installation process and added a section of configuration of openshift cluster.

## Motivation
https://github.com/integr8ly/tutorial-web-app/issues/410

## What
Fixed reference to customServerConfig.json in server.js. Updated READme to give more detail to the installation process and added a section of configuration of openshift cluster.

## Why
During start up the configuration of the provisioned services can be set using an environment variable or a configuration file 'customServerConfig.json'.  This feature was not working.

## How
I changed the file reference in server.js from '/etc/webapp/customConfigFile.json' to '/etc/webapp/customConfigFile.json', which . correctly referenced to file on startup.

## Verification Steps
Verification of environment variable configuration:
create a json file with required configuration and reference it on app startup 
user:/tutorial-web-app$ OPENSHIFT_HOST=master.example.openshiftworkshop.com:443 SERVER_EXTRA_CONFIG_FILE=~/configFile.json yarn start:dev

Verification using customServerConfig.json:
create tutorial-web-app/etc/webapp/customServerConfig.json with required configuration and start app.
user:/tutorial-web-app$ OPENSHIFT_HOST=master.example.openshiftworkshop.com:443  yarn start:dev

Using either of the steps above the result should be as follows:

screenshot 2019-02-20 at 16 14 16

## Checklist:

- [ ] Code has been tested locally by PR requester

## Progress

- [x] Finished task
- [ ] TODO

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -55,6 +55,9 @@ $ yarn install
 $ yarn start:dev
 ----
 
+IMPORTANT: The webapp will automatically open (http://localhost:3006) in your browser and watch for file changes.
+When running locally, the available services list is mocked, and service urls set via env vars.
+
 === Cluster configuration
 If you are running tutorial-web-app against your cluster, then you will need to configure your cluster to allow callback to your localhost:
 
@@ -72,13 +75,7 @@ redirectURIs:
 - http://localhost:3006
 
 ----
-----
-$ exit
-----
 
-
-IMPORTANT: The webapp will automatically open (http://localhost:3006) in your browser and watch for file changes.
-When running locally, the available services list is mocked, and service urls set via env vars.
 
 The webapp will also automatically load the walkthroughs from `../tutorial-web-app-walkthroughs/walkthroughs`.
 
--- a/README.adoc
+++ b/README.adoc
@@ -39,21 +39,63 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage 
 :numbered:
 === Clone the project
 
-First you need to clone the https://github.com/integr8ly/tutorial-web-app-walkthroughs[walkthroughs repository] next to the tutorial web app
+First you need to clone the https://github.com/integr8ly/tutorial-web-app-walkthroughs[walkthroughs repository] next to the tutorial web app:
 
 [source,shell]
 ----
-$ git clone https://github.com/integr8ly/tutorial-web-app-walkthroughs.git
+home:$ git clone https://github.com/integr8ly/tutorial-web-app-walkthroughs.git
+home:$ cd tutorial-web-app-walkthroughs
+home:~/tutorial-web-app-walkthroughs$ install yarn
+home:~/tutorial-web-app-walkthroughs$ cd ..
 ----
 
 === Install dependencies
-Then change to the webapp directory, install its dependencies and run the development server:
+Then change to the webapp directory, install its dependencies and run the development server.  
+[source,shell]
+----
+home:$ cd tutorial-web-app
+home:~/tutorial-web-app$ yarn install
+home:~/tutorial-web-app$ yarn start:dev
+----
+
+=== Cluster configuration
+If you are running tutorial-web-app against your cluster, then you will need to configure your cluster to allow callback to your localhost:
+
+
 
 [source,shell]
 ----
-$ yarn install
-$ yarn start:dev
+$ ssh -i ~/.ssh/ocp-workshop.pem user@bastion.example.openshiftworkshop.com
+[user@bastion ~]# sudo -s
+[root@bastion user]# oc edit oauthclient tutorial-web-app
 ----
+This will open the file in vim for editing. The line 'http://localhost:3006' should be added to the bottom of the oauthclient tutorial-web-app and the file saved
+----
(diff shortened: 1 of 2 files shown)


README.adoc
```diff
@@ -39,22 +39,64 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage
 :numbered:
 === Clone the project
 
-First you need to clone the https://github.com/integr8ly/tutorial-web-app-walkthroughs[walkthroughs repository] next to the tutorial web app
+First you need to clone the https://github.com/integr8ly/tutorial-web-app-walkthroughs[walkthroughs repository] next to the tutorial web app:
 
 [source,shell]
 ----
-$ git clone https://github.com/integr8ly/tutorial-web-app-walkthroughs.git
+home:$ git clone https://github.com/integr8ly/tutorial-web-app-walkthroughs.git
```
aidenkeating: I think we should remove the prompt part of this `home:*` so that a user can just copy the commands without thinking.

README.adoc
```diff
@@ -39,22 +39,64 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage
...
+home:$ cd tutorial-web-app-walkthroughs
+home:~/tutorial-web-app-walkthroughs$ install yarn
```
aidenkeating: Should this be `yarn install`? I wasn't aware we needed any dependencies in the `tutorial-web-app-walkthroughs` repo though. I can double check whether this is required

README.adoc
```diff
@@ -39,22 +39,64 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage
...
+=== Cluster configuration
+If you are running tutorial-web-app against your cluster, then you will need to configure your cluster to allow callback to your localhost:
+
+
 
 [source,shell]
 ----
-$ yarn install
-$ yarn start:dev
+$ ssh -i ~/.ssh/ocp-workshop.pem user@bastion.example.openshiftworkshop.com
```
aidenkeating: I don't think we should have any reference to hosts here. A user doesn't necessarily need to ssh onto the host, they could also login with the `oc` tool. maybe something like:

```
oc login <my-cluster>
oc edit oauthclient tutorial-web-app
```

and removing the `exit` further down the file would be more generic

README.adoc
```diff
@@ -39,22 +39,64 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage
...
+grantMethod: auto
+kind: OAuthClient
+metadata:
+  creationTimestamp: 2019-02-25T13:11:51Z
+  name: tutorial-web-app
+  resourceVersion: "23196"
+  selfLink: /apis/oauth.openshift.io/v1/oauthclients/tutorial-web-app
+  uid: ea3dc674-38fe-11e9-9ca1-0601b82a88d2
+redirectURIs:
+- https://tutorial-web-app-webapp.apps.tonydavidson-750a.openshiftworkshop.com
```
aidenkeating: Let's make this more generic to remove any reference to specific hosts. Maybe:

```
redirectURIs:
- ...
- http://localhost:3006
```

to signify they should add it to the end of the list? wdyt?

README.adoc
```diff
@@ -39,22 +39,64 @@ TIP: Use https://github.com/creationix/nvm/blob/master/README.md[nvm] to manage
...
+- http://localhost:3006
+~
```
aidenkeating: We can remove from here down, seems to be output from vim.

server.js
```diff
@@ -15,7 +15,7 @@ const app = express();
 app.use(bodyParser.json());
 
 const port = process.env.PORT || 5001;
-const configPath = process.env.SERVER_EXTRA_CONFIG_FILE || '/etc/webapp/customServerConfig.json';
+const configPath = process.env.SERVER_EXTRA_CONFIG_FILE || './etc/webapp/customServerConfig.json';
```
aidenkeating: This would always point to a relative path inside the repo, I don't think we want this. The current main use of this env var places it's config in the path `/etc/webapp/customServerConfig.json` so I think we're good to keep this the same.

If a user wants to use the relative path instead, they can still use the `SERVER_EXTRA_CONFIG_FILE ` env

README.adoc
tonyxrmdavidson: Yeah, that makes sense.

README.adoc
tonyxrmdavidson: I just did the yarn install as a matter of habit.  I didn't actually test it without doing this. If you have done it previously without then best to take it out.

README.adoc
tonyxrmdavidson: Yes, that would be better.

README.adoc
tonyxrmdavidson: Yep . makes sense.

README.adoc
aidenkeating: Confirmed that a `yarn install` isn't required here.

server.js
tonyxrmdavidson: OK,  that makes sense, but at least we know the code works now.

PR #436, opened 2019-03-06
INTLY-1063: Use a supported base container image for the webapp

mfrances17:
## Motivation
INTLY-1063 - Use a supported base container image for the webapp

## What
Replaced the node.js version in the FROM with a supported rh version. Needed to specify a new path LD_LIBRARY_PATH to fix docker build issues.

## Why
So the webapp is running on a supported node.js & rhel version, and can take advantage of CVE notifications.

## Verification Steps
1. Verify that the docker build of the webapp succeeds without errors.
2. Verify that when you point the webapp to the docker build image, it looks and behaves as expected.

## Checklist:

## Progress
- [x] Finished task

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,7 +13,7 @@ COPY . ./
 
 USER root
 
-RUN yum -y install git
+RUN yum -y install git httpd24-libcurl
 
 RUN chmod g+w yarn.lock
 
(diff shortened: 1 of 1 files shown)


pb82:
:eyes:

PR #442, opened 2019-03-15
Add about endpoint to retrieve configured walkthrough repos

aidenkeating:
Adds an `/about` endpoint which returns the version of the webapp along with the configured walkthrough location repos and commits.

An example response would be something like:

```
{
  "version":"2.5.0",
  "walkthroughLocations": [
    {
      "type": "git",
      "commitHash": "a01cc3159cad876ce41f1c71aa21e8a4bc5b7726",
      "commitDate": "2019-03-08 17:12:10 -0800",
      "remote": "https://github.com/RedHatWorkshops/dayinthelife-integration",
      "folder":"docs/labs/developer-track/"
    },
    ....
   ]
}
```

### DIFF
--- a/git_client.js
+++ b/git_client.js
@@ -1,4 +1,4 @@
-const simpleGit = require('simple-git/promise')(__dirname);
+const simpleGit = require('simple-git/promise');
 const path = require('path');
 const url = require('url');
 const fs = require('fs');
@@ -10,7 +10,7 @@ function getRepoName(repoUrl) {
 
 /**
  * Retrieve the base git repo URL from a provided URL.
- * 
+ *
  * A URL could contain a hash, which we're interpreting as a branch/tag separator,
  * we shouldn't include this.
  **/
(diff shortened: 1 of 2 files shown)


tiffanynolan:
If this would be a good location to do so, is it possible to include some type of "pretty name" data for the walkthrough locations here as well, that could be used on the dashboard as a heading?  
"prettyName": "Developer Track"

mfrances17:
> If this would be a good location to do so, is it possible to include some type of "pretty name" data for the walkthrough locations here as well, that could be used on the dashboard as a heading?
> "prettyName": "Developer Track"
> 
> FYI @mfrances17

@tiffanynolan Currently, I just use 'Red Hat' if it's one of our GitHub repos, and 'Custom' if not. I didn't think there was a  requirement to do anything more than that, unless I missed something in the mocks (quite possible)?

tiffanynolan:
The pretty name would be for the overall display heading for that group of walkthroughs on the dashboard. It's in addition to the Link text and URL for the repo that would be used in the metadata panel, as this data would be used on the dashboard as well. This might not be the right place to put it. I'll setup a quick meeting to discuss more.

PR #451, opened 2019-04-03
INTLY-1362 Use managedServices in walkthrough manifest

aidenkeating:
Currently, we provision "slices" of AMQ Online and Fuse Online when
a user logs in for the first time. However, a user may never use a
provisioned "slice", meaning resources are being wasted on a cluster.

This change allows these "slices" to be defined in the Walkthrough
configuration itself, so that they will be provisioned as-neeeded.

This should help with resource usages in the cluster.

Verification:
- Run the server, pointing at an Integreatly cluster from `master`
and with `WALKTHROUGH_LOCATIONS` pointing to
`https://github.com/aidenkeating/tutorial-web-app-walkthroughs.git#INTLY-1362-add-shared-services-config`
- [x] Ensure that on first login, AMQ Online and the non-Shared Fuse
Online are no longer being provisioned (check the namespace).
- [x] Ensure every service except AMQ Online on the right-hand side is clickable and the dashboards show as expected
- [x] Ensure that AMQ Online is visible in the right-hand bar of the landing page with a `Launch` button.
- [x] Click the `Launch` button, ensure the button disappears and the service starts provisioning
- [x] Ensure AMQ Online eventually hits a provisioned state in the landing page
- Click on Walkthrough 1A
- [x] Ensure the loading screen has the same layout as the initial loading screen (see images in comments).
- [x] Ensure the loading screen stays visible
until the AMQ Online and Fuse Onlne slices have been fully provisioned
and the NodeJS Application is provided with the correct credentials.
- Load another Walkthrough, ensure they still load as expected without issues.


aidenkeating:
Seeing an issue with the AMQ Online URL on the Walkthrough side-bar. Fixing that now

aidenkeating:
The related Walkthroughs PR (https://github.com/integr8ly/tutorial-web-app-walkthroughs/pull/62)

david-martin:
Unusual behaviour when i run the webapp locally pointed at the same cluster.
The amq online and fuse application links don't show on the right
image

I'll look into this, and the product name not getting replaced

src/components/loadingScreen/loadingScreen.js
```diff
@@ -52,12 +47,10 @@ LoadingScreen.propTypes = {
   staticThrobberImage: PropTypes.string,
   throbberImage: PropTypes.string,
   loadingText: PropTypes.string,
-  standbyText: PropTypes.string,
-  hideDelay: PropTypes.number
+  standbyText: PropTypes.string
 };
 
 LoadingScreen.defaultProps = {
-  hideDelay: 2500,
```
david-martin: @aidenkeating what is the reason for these changes to the loading screen?
Are these cleanup changes as the hideDelay was never used?

tiffanynolan:
@david-martin @aidenkeating - I created a new JIRA issue to update the walkthrough loading screen. I'm hoping we get to this next sprint (starting later today). Feel free to add any info to the task in JIRA that might be relevant. https://issues.jboss.org/browse/INTLY-1659

david-martin:
@aidenkeating I've pushed a small update to provision fuse & amq online into the shared user namespace when starting w1a.

pb82:
@david-martin The issues you described (non-pretty names in the app list, both fuse and amq-online in the apps list) can be reproduced when the product details cannot be looked up here: https://github.com/integr8ly/tutorial-web-app/blob/master/src/components/installedAppsView/InstalledAppsView.js#L23

The product details are defined in this file: https://github.com/integr8ly/tutorial-web-app/blob/master/src/product-info.js

So either the file cannot be located or the app names are different for some reason. Doing more investigation. Were you able to reproduce this locally or only on a cluster? Did that happen when you navigated back to the Overview from a Walkthrough?

src/components/loadingScreen/loadingScreen.js
aidenkeating: @david-martin This was due to a change in the way the loading screen is displayed. In `task.js` it's now the default state of the page and will be overridden with the walkthrough when everything is ready. So it's just taking the logic for when the loading screen appears and disappears out of the LoadingScreen component itself and into the implementation. Happy to revert it though.

aidenkeating:
@david-martin That was a mistake on my part. I've pushed up linting fixes for the commit. What is the expected behaviour on the home screen? Do we want the services to be shown but not-enabled by default on the right? Or are we ok with them popping up afterwards?

aidenkeating:
@david-martin I managed to reproduce the issue you were seeing, it was down to the way product details were being retrieved. I've also resolved the AMQ Online on the landing page vanishing.

This is the commit: https://github.com/integr8ly/tutorial-web-app/commit/c1fd6588a82cfbef98796e3a48eba4b794047442

tiffanynolan:
@aidenkeating 

> The only concern I have with this is we won't really have a way to tell the user how to provision the service without hardcoding in some message about 1A, which is something we've tried to avoid previously to allow the webapp to be flexible. So they'll see AMQ Online isn't available and won't know why.

Can the user initiate the provisioning from the Applications list as well if they select that item in the data list, or is the ONLY way by launching 1A?

aidenkeating:
@tiffanynolan 1A is currently the only way of provisioning AMQ Online for a user.

Something like being able to press the disabled AMQ Online button to provision it if the user doesn't want to go into 1A does seem like a good idea to me though. Not sure how that would look from a UX point of view, but it shouldn't be too much work engineering-wise.

Do we want to hold off on this PR merging until we've resolved this part? Or are we fine with this being merged in the meantime?

tiffanynolan:
@aidenkeating I added some comments and design specs to the Jira. My preference would be to not merge this until it's more obvious to the user as to what's happening in the Applications list.

aidenkeating:
@tiffanynolan Mind taking a look? I've included examples of the loading screen and launch buttons below.

Loading screen was added in: b6fec9f
Old loading screen was removed in: 768b475
The launch button was added in: 2b64cea

## The per-walkthrough loading screen

This only shows the additional services that have been added by the user to the `managedServices` section of their walkthrough JSON configuration. So for 1A that's AMQ Online and Fuse Online
Screen Shot 2019-04-11 at 10 42 58

### The landing page launch button

At the moment it's only used for AMQ Online, but a pattern has been set up to allow others to be added easily too.
launch-amq-online

aidenkeating:
As the changes in this PR are quite extensive I'm going to go through and test less used features like custom services and hiding services to ensure nothing has been broken.

aidenkeating:
Added one more small change to keep the custom server config working as expected. If you override ther services provision to just be `3scale` and `amq-online-standard` you'll see the following when starting the web app:

Screen Shot 2019-04-11 at 11 22 39

When you update the new option in the custom server config to override the services to display even when they're not available to `[]` you'll see the following on the right-hand list:

Screen Shot 2019-04-11 at 12 13 53

aidenkeating:
@david-martin @pb82 I've updated verification steps for this PR to include loading screen and launch button changes suggested by @tiffanynolan

tiffanynolan:
@aidenkeating This is looking great. Thanks for making the changes. I'll take a closer look later today but did notice one thing right away. Can you change the link from "Launch" to "Start service"?

aidenkeating:
@tiffanynolan Updated.

Screen Shot 2019-04-11 at 17 22 35

@david-martin @pb82 I've deployed this to `akeating-2df8`. The cluster seems pretty slow. I may reprovision. Will keep this comment updated.

tiffanynolan:
UX-wise this looks great! Thanks for making the changes, @aidenkeating. 

I just noticed the "Community" label is still being displayed with Red Hat Developer Launcher in the Applications list. If you are already editing the file where you can remove this as part of this PR, can you remove that label too? It should not have the Community label associates with it as it is a Red Hat "product". If not, let me know and I'll create a new Jira to have that label removed.

aidenkeating:
@tiffanynolan Thanks for the review. I've updated the web app so the `community` tag will not be shown beside Launcher

aidenkeating:
@david-martin @pb82 @maleck13 @odra I think this is now ready for verification.

Feel free to use my cluster, which already has the PR set up: `akeating-a7d6`
Or here's an image of the web app in this PR: `docker.io/aidenkeatingrht/tutorial-web-app:INTLY-1362`

matskiv:
Verifying on 'akeating-a7d6' cluster with evals40 user.

matskiv:
@aidenkeating I am seeing blank screen when I open W1A. This is in the console:
'''RangeError: invalid date[Learn More] walkthroughDetails.js:42:63'''

PR #452, opened 2019-04-03
add rhsso default service

pb82:
Adds a link to RH-SSO to the applications list

Verification steps:

This and https://github.com/integr8ly/managed-service-broker/pull/37 are deployed to https://master.pbraun-73b1.openshiftworkshop.com/console/catalog

1. Login to the webapp on this cluster using an evals user (evals12 is taken)
1. Wait until all services are provisioned (there should be an entry for Red Hat Single Sign-On in the list)
1. Make sure that the application list on the overview page has an entry for Red Hat Single Sign-On
1. Make sure that the link is correct


tiffanynolan:
@pb82 @finp For consistency with the other products in the list where we're using the full product name, I'd recommend that we use "Red Hat Single Sign-On" as the prettyName in the product-info file.

pb82:
@tiffanynolan Updated

david-martin:
@pb82 Tested & verified OK.
In your cluster, you set the env var for the broker to the root of SSO.
Lets set this to the openshift realm login page when setting that env var in the installer

PR #456, opened 2019-04-23
INTLY-1814 Don't crash when walkthrough isn't found

aidenkeating:
Currently the sever will crash if a user tries to retrieve a walkthrough that doesn't exist. Although this should be rare, this resolves both the UI and server from completely failing.

@tremes I'm building an image now for testing. Verification steps are below. 

@tiffanynolan Do we have any designs for a not found screen that could be used on the UI in the future? Is this something we care too much about?

## Verification steps
- Use the image of the webapp, `docker.io/aidenkeatingrht/tutorial-web-app:INTLY-1814`
- Go to a route for a walkthrough that doesn't exist, for example `<webapp-url>/tutorial/test`
- Ensure the server doesn't crash and prints an error message instead.
- Ensure the UI doesn't completely crash, it should just be a blank screen.

### DIFF
--- a/server.js
+++ b/server.js
@@ -130,16 +130,17 @@ app.get('/about', (_, res) => {
 
 app.get('/about/walkthrough/:walkthroughId', (req, res) => {
   const { walkthroughId } = req.params;
+  const walkthrough = walkthroughs.find(w => w.id === walkthroughId);
+  if (!walkthrough) {
+    console.error('Could not find walkthrough with ID', walkthroughId);
+    res.sendStatus(404);
+    return;
+  }
   res.json({
     walkthroughId,
-    walkthroughLocation: getWalkthroughLocationInfoForWalkthrough(walkthroughId)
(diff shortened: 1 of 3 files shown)


tremes:
@aidenkeating Thanks. It looks good to me. Btw good question. I think we should handle 404-like screens somehow.

tiffanynolan:
@aidenkeating @tremes Good catch on this. I didn't know we didn't have something in place for 404s, 500s etc. I know we have error screen designs for other projects so we should be able to come up with a UI to display for these fairly easily. I'll create an item in our backlog for it so it can be prioritized from the design side.

mfrances17:
Code looks good. Just going to do some testing before approving.

pb82:
@mfrances17 the difference is that the webapp backend no longer crashes with `TypeError` but instead just prints a message `Could not find walkthrough with ID tutorial-web-app-walkthroughs-walkthroughs-3-fuse-aggregator-and-api-management`

I think this makes it easier to go through the logs.

PR #462, opened 2019-05-01
switch to ns

odra:
uses openshift namespace an enmasse provisioning services "namespace" param


maleck13:
verified on my cluster. Could we merge and create a new release

PR #467, opened 2019-05-27
Intly 1993

pb82:
Adds endpoints to the webapp to manage user defined walkthrough repos.

Verification steps:

* Start the webapp locally
* Add a new user defined repo with curl: `curl -XPOST localhost:3006/user_walkthroughs -d '{"data": "https://github.com/integr8ly/example-customisations.git"}' -H "Content-Type: application/json"`
* Resync the walkthroughs: `curl -XPOST localhost:3006/sync-walkthroughs`
* Refresh the browser: you should now see the extra walkthroughs from the added repository.
* Run `curl -XPOST localhost:3006/user_walkthroughs -d '{"data": ""}' -H "Content-Type: application/json"`
* Run `curl -XPOST localhost:3006/sync-walkthroughs`
* Refresh the browser again. The extra WTs should be gone now.
* You can also try adding multiple repositories in `data` separated by "\n" chars.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -36,3 +36,6 @@ yarn-error.log*
 
 # ignore package-lock file, as the webapp uses yarn
 package-lock.json
+
+# Ignore database files
+*.db
--- a/model/database.js
+++ b/model/database.js
@@ -1,0 +1,10 @@
+const { Sequelize } = require('sequelize');
+const { join } = require('path');
+
+exports.createDatabase = (dbPath, dbName) => {
+  return new Sequelize(`${dbName}`, null, null, {
+    storage: join(__dirname, dbPath, `${dbName}.db`),
+    dialect: 'sqlite',
+    logging: true
+  });
+};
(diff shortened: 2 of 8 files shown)


pb82:
@mfrances17 are you happy with those endpoints for your use case?

mfrances17:
I'm not sure about the delete use case. I was planning on GET and POST of course. I see that you pass an ID of 1 - where is that ID assigned to the user-defined repo? Is it just the order they appear in the textarea? They can rearrange the order. What I was planning to do was do a GET on load of the settings page and a full DELETE and then POST when they click Save. Will I be able to do that with these endpoints - is there a 'delete all'?

pb82:
@mfrances17 The ID is returned as part of the repository (id, url). When you list all repositories you get an array of objects, each with id and url. When you create a new repository you get one object (the created one) back with id and url.

pb82:
@mfrances17 I've updated the endpoints now to store all URLs in one row. A POST to `/user_walkthroughs` with a body of 

```json
{"data": "<user walkthroughs separated by newlines>"}
```

will udpate the value. To remove all walkthrougs, simply set `data` to an empty string.

PR #474, opened 2019-06-26
INTLY-1285: Custom walkthroughs via Solution Explorer settings UI

mfrances17:
## Motivation
https://issues.jboss.org/browse/INTLY-1285

## What
Add a new Settings UI to the Solution Explorer which allows users to add github URLs to their own walkthrough content. On Save, the UI will show valid walkthroughs in the same way that editing the old walkthroughs environment variable did.

## Why
Make it easier for users to add custom walkthroughs without going to the OpenShift environment variables.

## Verification Steps
General steps:
1) Click the new Settings cog in the masthead.

2) Type or paste the above repo into the ‘List URLs…’ text area and click Save. You’ll bounce to the Home page automatically. Here’s a github URL you can use as a custom repo addition:
https://www.github.com/integr8ly/example-customisations

3) Refresh your browser. Walkthroughs from the entered repo will now be listed on the Home page.

4) Click the Settings cog.

5) Clear the ‘List URLs…’ text area and click Save, which will bring you back to the Home page.

6) Refresh your browser. Walkthroughs from the custom repo will be removed from the Home page.

## Checklist:

- [x] Code has been tested locally and on a live server by PR requester

## Progress
- [x] Finished task

## Additional Notes
The Home page will not list any repos if you do any of the following:
• List the same custom repo twice
• Enter a valid github URL but does not contain valid walkthrough configuration
Errors for these situations display in the OpenShift logs, just as they do now when doing either of the above using the environment variable.

You are free to use my server to test the changes, if you want:
https://tutorial-web-app-webapp.apps.uxddev-0c8d.openshiftworkshop.com

Or if you prefer, you can also deploy this webapp image on your own OpenShift instance:
docker.io/mfrances17/dev-tutorial-web-app:latest

Screen cap of new Settings page:
Screen Shot 2019-06-26 at 2 18 52 PM


src/components/masthead/masthead.js
```diff
@@ -86,6 +92,16 @@ class Masthead extends React.Component {
...
         <Toolbar>
           <ToolbarGroup className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnLg)}>
+            <ToolbarItem className={css(accessibleStyles.screenReader, accessibleStyles.visibleOnMd)}>
+              <Button
+                className="pf-c-button pf-m-plain"
+                aria-label="Settings"
+                variant="plain"
+                onClick={this.onSettingsClick}
+              >
+                <i className="fas fa-cog" aria-hidden="true" />
```
mindreeper2420: I think there is a React Icon for this, but we can update in a future story.

dlabaj: yeah you can import the Cog icon
```suggestion
                < CogIcon/>
```

then just add an import above as 

```suggestion
                import { CogIcon } from '@patternfly/react-icons';
```

src/pages/settings/settings.js
```diff
@@ -0,0 +1,200 @@
...
+  };
+
+  handleTextInputChange = value => {
+    this.setState(
+      {
+        value,
+        isValid: /^(?:https:\/\/)+([w.-]+)+github.com\/[\w\-._~:/?#[\]@!$&/'()*+,;=.]+$/.test(value)
+      },
+      () => {
+        if (this.state.value === '') {
```
dlabaj: could value ever be undefined or null?

src/pages/settings/settings.js
```diff
@@ -0,0 +1,200 @@
...
+        }
+      }
+    );
+  };
+
+  render() {
+    const { value, isValid } = this.state;
+
+    return (
+      <React.Fragment>
```
dlabaj: I don't think you need React.Fragment here

pb82:
@mfrances17 When i add `https://www.github.com/integr8ly/tutorial-web-app-walkthroughs` to the list and then refresh the page, all walkthroughs disappear.

Also it seems like the value is only accepted if there is a newline after the last entry. Maybe we should mention this on the settings page?

src/pages/settings/settings.js
mfrances17: Yep, I removed it, works fine without.

mfrances17:
> @mfrances17 When i add `https://www.github.com/integr8ly/tutorial-web-app-walkthroughs` to the list and then refresh the page, all walkthroughs disappear.
@pb82 I think that's because it's a duplicate of the default walkthroughs repo, it will fail and show an error in the log on OpenShift.

> Also it seems like the value is only accepted if there is a newline after the last entry. Maybe we should mention this on the settings page?
I will test it again, I thought it was working without needing this.

src/pages/settings/settings.js
mfrances17: If the field is empty is will be "" and otherwise have some value, which if it isn't a valid github formatted URL, the Save button will be inactive and they won't be able to write it to the db.

src/components/masthead/masthead.js
mfrances17: I imported the CogIcon from react-icons and it looks good. Thanks.

mfrances17:
> > @mfrances17 When i add `https://www.github.com/integr8ly/tutorial-web-app-walkthroughs` to the list and then refresh the page, all walkthroughs disappear.
> > @pb82 I think that's because it's a duplicate of the default walkthroughs repo, it will fail and show an error in the log on OpenShift.
> 
> > Also it seems like the value is only accepted if there is a newline after the last entry. Maybe we should mention this on the settings page?
> > I will test it again, I thought it was working without needing this.

Update: I re-tested using the following repositories:

https://www.github.com/integr8ly/example-customisations 
https://www.github.com/finp/walk-about

Both repositories' walkthroughs displayed on the Home page, in addition to the default repos, without requiring a newline char after the last entry.

PR #481, opened 2019-07-11
Close connections

pb82:
## Motivation

Close the connection to the sqlite database and also stop the webserver before shutting down the webapp.


mfrances17:
Can you provide some quick steps or recommendations for testing that the connections are actually closed? I'd like to verify on my server.

pb82:
@mfrances17 sorry, i missed your comment. The best way to test this is actually to run the webapp locally using `node server.js`, then kill it with Ctrl+C: you should see the additional log output for shutting down. When using yarn those logs are suppressed.

PR #488, opened 2019-07-24
Intly 2588

pb82:
Make `clusterType` (RHPDS, ODS...) available to the webapp. This needs an operator and installer change to actually set the value.

Verification steps in https://github.com/integr8ly/installation/pull/766


pb82:
Verified in https://github.com/integr8ly/installation/pull/766

PR #490, opened 2019-07-29
INTLY-2347 Don't use Service Instance for AMQ Online

aidenkeating:
Verification:
- Run the server locally against an Integreatly OpenShift 3 cluster
  - `OPENSHIFT_HOST=<openshift-host> SSO_ROUTE=<sso-route> GITEA_TOKEN=<gitea-token> GITEA_HOST=<gitea-host> yarn start:dev`
  - You can get the values for the above command with with `oc get dc tutorial-web-app -n webapp -o yaml`
- Wait for `localhost:3006` to load in your browser
- Login as `evals01`, for example
- Start walkthrough 1 (AMQ Online and Fuse Online)
- Ensure AMQ Online is provisioned and shown in the loading screen
  - `oc get addressspace --all-namespaces` has an AddressSpace named after the user
  - `oc get messaginguser --all-namespaces` has a MessagingUser in the created AddressSpace
- Go through Walkthrough 1 and ensure it still works as expected
- Login as a different user, provision AMQ Online from the right-hand side of the main page


aidenkeating:
@leaannegaffney @dimitraz Mind taking a look?

dimitraz:
:eyes:

PR #493, opened 2019-08-01
INTLY-2663 add user sso provision to startup loading task

aidenkeating:
verification:
- an integreatly cluster on openshift 3 is required
- ensure the user-rhsso clusterserviceclass exists by searching for
  it on the main page (service catalog)
- ensure USER_SSO_URL is set on the managed service broker env
  var
- run this branch locally against the cluster
- ensure that the user sso is now also provisioned at startup
- ensure the link on the right-hand side works as expected


tiffanynolan:
@aidenkeating Do you have a cluster running with this already that you could share? I'd like to verify the display of this in SolX.

aidenkeating:
@tiffanynolan I will set up an environment now for verification

aidenkeating:
@tiffanynolan @leaannegaffney This is provisioned on my cluster here https://tutorial-web-app-webapp.apps.akeating-242a.openshiftworkshop.com/, the `admin` user has been set up already

src/product-info.js
```diff
@@ -32,6 +32,10 @@ export default {
     gaStatus: 'GA'
   },
   rhsso: {
+    prettyName: 'Red Hat Single Sign-On (Cluster)',
```
finp: @tiffanynolan this doesn't align with doc at https://integr8ly.github.io/user-documentation/index.html?e=docs/getting-started.adoc#gs-adding-users-proc

but we can adjust as required

pb82:
@aidenkeating can you also add the user sso to the mock data here: https://github.com/integr8ly/tutorial-web-app/blob/master/server.js#L495

otherwise running the webapp locally won't work

src/product-info.js
tiffanynolan: @finp I like how you referred to the 2 instances in the docs as those terms sound more user friendly. Ultimately, I'd like to only show the end-user SSO in the applications list and move the cluster level one to more of an admin screen/settings page. We're not there yet though so we do need some way to easily tell them apart for now.

aidenkeating: @tiffanynolan @finp Just to confirm, should I be changing the naming of this in the web app (and if so, what to)? Or will it be changed in the documentation?

finp: @aidenkeating my reading of  @tiffanynolan comment is to go with what we have. I could add a note to docs if required.

aidenkeating:
@leaannegaffney Mind taking a look?

PR #494, opened 2019-08-02
feat: Add mdc url helper

wtrocki:
So I have been trying to add MDC (mobile developer console) URL into the tutorial-web-app.

I see service available in config map

Screenshot 2019-08-02 at 13 19 38

SO I have created new type in DEFAULT_SERVICES and added new helper `mdc-url`.
Looking to verify that now but positing for early feedback.

## Verification

* Build an image with this pr or use `docker.io/odranoel/webapp:dev`
* Change the webapp deployment config to use this image
* Set both env vars `OPTIONAL_WATCH_SERVICES` and `OPTIONAL_PROVISION_SERVICES` to "mdc" and wait for the pod to be ready

### DIFF
--- a/server.js
+++ b/server.js
@@ -620,8 +620,28 @@ function getConfigData(req) {
     }/auth/realms/openshift/protocol/openid-connect/logout?redirect_uri=${logoutRedirectUri}',
     threescaleWildcardDomain: '${process.env.THREESCALE_WILDCARD_DOMAIN || ''}',
     integreatlyVersion: '${process.env.INTEGREATLY_VERSION || ''}',
-    clusterType: '${process.env.CLUSTER_TYPE || ''}'
+    clusterType: '${process.env.CLUSTER_TYPE || ''}',
+    optionalWatchServices: ${arrayFromString(process.env.OPTIONAL_WATCH_SERVICES, ',')},
+    optionalProvisionServices: ${arrayFromString(process.env.OPTIONAL_PROVISION_SERVICES, ',')}
   };`;
+}
+
+function arrayFromString(data, sep) {
+  if (!data) {
+    return '[]';
+  }
+
+  let parts = ['['];
+  data.split(sep).forEach(function(v, idx, arr) {
+    if (idx === arr.length -1) {
+        parts.push(`"${v}"`);    
+    } else {
+        parts.push(`"${v}",`);
+    }
+  });
+  parts.push(']');
+
+  return parts.join('');
 }
 
 function getWalkthroughInfoFromAdoc(parentId, id, dirName, doc) {
--- a/server.js
+++ b/server.js
@@ -621,28 +621,12 @@ function getConfigData(req) {
     threescaleWildcardDomain: '${process.env.THREESCALE_WILDCARD_DOMAIN || ''}',
     integreatlyVersion: '${process.env.INTEGREATLY_VERSION || ''}',
     clusterType: '${process.env.CLUSTER_TYPE || ''}',
-    optionalWatchServices: ${arrayFromString(process.env.OPTIONAL_WATCH_SERVICES, ',')},
-    optionalProvisionServices: ${arrayFromString(process.env.OPTIONAL_PROVISION_SERVICES, ',')}
+    optionalWatchServices: ${JSON.stringify(arrayFromString(process.env.OPTIONAL_WATCH_SERVICES || '', ','))},
+    optionalProvisionServices: ${JSON.stringify(arrayFromString(process.env.OPTIONAL_PROVISION_SERVICES || '', ','))}
   };`;
 }
 
-function arrayFromString(data, sep) {
-  if (!data) {
-    return '[]';
-  }
-
-  let parts = ['['];
-  data.split(sep).forEach(function(v, idx, arr) {
-    if (idx === arr.length -1) {
-        parts.push(`"${v}"`);    
-    } else {
-        parts.push(`"${v}",`);
-    }
-  });
-  parts.push(']');
-
-  return parts.join('');
-}
+const arrayFromString = (data, sep) => data.split(sep).filter(item => item !== '');
 
 function getWalkthroughInfoFromAdoc(parentId, id, dirName, doc) {
   // Retrieve the short description. There must be a gap between the document title and the short description.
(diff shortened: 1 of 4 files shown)


src/common/serviceInstanceHelpers.js
```diff
@@ -60,7 +60,8 @@ const DEFAULT_SERVICES = {
   THREESCALE: '3scale',
   APICURIO: 'apicurio',
   FUSE_MANAGED: 'fuse-managed',
-  RHSSO: 'rhsso'
+  RHSSO: 'rhsso',
+  MDC: 'Mobile Developer Console'
```
wtrocki: Not sure what is the key and where it comes from. I do not see rhsso in config map so it is probably created somewhere else?

wtrocki: How I can add mdc to it and use URL from the secret?

wtrocki:
Ok so this definietly is not going to work. Any help very appreciated.

odra:
@pb82 one question: I thought that a service needed to be added in msb for the webapp to list it, is that correct?

pb82:
@odra Not only to list it, but the webapp relies on the service instance (created by the MSB) to lookup the dashboard URL for this service.

wtrocki:
I have found this nice class so it looks that this secret is not used and we need to do something extra: https://github.com/integr8ly/tutorial-web-app/blob/81c5741c2387f0fe482ae0501593a8fa7082102e/src/services/middlewareServices.js

wtrocki:
Replicated changes as in https://github.com/integr8ly/tutorial-web-app/pull/493

Thx @aidenkeating :)

> @odra Not only to list it, but the webapp relies on the service instance (created by the MSB) to lookup the dashboard URL for this service.

To be clear: we need to replicate changes made in https://github.com/integr8ly/managed-service-broker/pull/45/files for MDC?

wtrocki:
Fixed unit tests

odra:
That's what we need to add it in msb, I just wanted to know if is required to do so at this point

src/services/middlewareServices.js
```diff
@@ -48,7 +49,8 @@ const PROVISION_SERVICES = [
   DEFAULT_SERVICES.APICURIO,
   DEFAULT_SERVICES.THREESCALE,
   DEFAULT_SERVICES.FUSE_MANAGED,
-  DEFAULT_SERVICES.RHSSO
+  DEFAULT_SERVICES.RHSSO,
+  DEFAULT_SERVICES.MDC
```
aidenkeating: Will this not result in the webapp trying to provision the MDC every time, regardless of whether it exists or not? From what I understood the MDC is an optional add-on in the current installer and may not be found. Is that still correct?

wtrocki: @aidenkeating I have very limited knowledge of the webapp. 
Our target is only to *get url of the MDC* for our walkthorugh. 
MDC will be always installed and do not need to be proisioned. 
If you know any secret on how we can get access to MDC url god to bless you. What we have is MDC server running in mdc namespace with some route.

I see URL available in Application manifest (see above) in MDC so Web app already knows this url, but I personally struggle to get easy way to expose this url to walkthough. Maybe we can have something custom for MDC that will be able to consume value above?


@wei-lee correct me if I'm wrong.

odra:
@wtrocki is this pr ready for verification? I  already added mdc in msb and wanted to test it

src/services/middlewareServices.js
odra: mdc will be installed by default now (once we merge the feature branch)

odra: I am making it optional and will update this pr with my commit as soon as I am done with it

wtrocki:
@odra Yes. We can use this or change walktrhough.json in mobile-walkthorugh to use shared resource. I haven't tried it but this change replicates other PR's so this technically should work

odra:
@aidenkeating  @pb82 could please give me your review/opinion about the changes again?

server.js
```diff
@@ -620,10 +620,30 @@ function getConfigData(req) {
     }/auth/realms/openshift/protocol/openid-connect/logout?redirect_uri=${logoutRedirectUri}',
     threescaleWildcardDomain: '${process.env.THREESCALE_WILDCARD_DOMAIN || ''}',
     integreatlyVersion: '${process.env.INTEGREATLY_VERSION || ''}',
-    clusterType: '${process.env.CLUSTER_TYPE || ''}'
+    clusterType: '${process.env.CLUSTER_TYPE || ''}',
+    optionalWatchServices: ${arrayFromString(process.env.OPTIONAL_WATCH_SERVICES, ',')},
+    optionalProvisionServices: ${arrayFromString(process.env.OPTIONAL_PROVISION_SERVICES, ',')}
```
odra: both those vars should hold a comma separated value of additonal services - I decided to use two arrays so it would be easiet to set it in ansible

server.js
```diff
@@ -620,10 +620,30 @@ function getConfigData(req) {
...
   };`;
 }
 
+function arrayFromString(data, sep) {
```
odra: I had to write this function since the value needs to be a valid array in string formatting

src/services/middlewareServices.js
```diff
@@ -52,6 +56,10 @@ const PROVISION_SERVICES = [
   DEFAULT_SERVICES.RHSSO,
   DEFAULT_SERVICES.USER_RHSSO
 ];
+if (window.OPENSHIFT_CONFIG && window.OPENSHIFT_CONFIG.optionalProvisionServices.length > 0) {
```
odra: it just appends the services into the existing array (well actually concats it) is there are optional services available

wei-lee:
I have verified this PR as part of https://github.com/integr8ly/managed-service-broker/pull/47

server.js
pb82: Why not something like:

```js
const arrayFromString = (data, sep) => data.split(sep).filter(item => item !== '');
```

odra: would that parse the array in string format? I got some problems using an array object within that formatted string

pb82: yes, if the string is in a format like `a,b`. An empty string would return an empty array.

odra: I mean `['mdc']`  was being parsed as "mdc", let me give it  a try again

PR #504, opened 2019-08-26
INTLY-3044 apicurio product info

aidenkeating:
currently, for the product apicurio, we are displaying the display
name as apicurito. this is now incorrect and we should instead be
using api manager.

this change replaces the display name for apicurio with api
designer.

verification:
- run yarn start:dev
- open the main page of the solution explorer
- ensure the apicurio service is now api designer
- if running against a cluster, ensure the link still works as
  expected

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreatly-web-app",
-  "version": "2.18.2",
+  "version": "2.18.3",
   "private": true,
   "proxy": "http://localhost:5001/",
   "dependencies": {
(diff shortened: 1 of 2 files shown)


aidenkeating:
Screen Shot 2019-08-26 at 23 38 35

@pb82 @tiffanynolan Mind taking a look?

pb82:
@aidenkeating API Designer or API Manager (because in the PR description its API Manager)? Otherwise LGTM.

PR #522, opened 2019-10-17
Default installed services

aidenkeating:
master on openshift 3 is failing due to the JSON returned being invalid when INSTALLED_SERVICES env isn't set.

verification steps:
- run the webapp from master, window shouldn't load
- run the webapp from this branch, window should load and provisioning screen should start

@JameelB is it correct that this value is only relevant in openshift 4, or should it be used in 3 also? are you happy with this default (an array) if nothing is set or should it be an object? i'm not overly familiar with this stuff

@tonyxrmdavidson mind taking a look, this should resolve the issue you're seeing

### DIFF
--- a/server.js
+++ b/server.js
@@ -642,6 +642,7 @@ function getConfigData(req) {
   const masterUri = isOpenShift4() ? OPENSHIFT_PROXY_PATH : `https://${process.env.OPENSHIFT_HOST}`;
   const wssMasterUri = isOpenShift4() ? OPENSHIFT_PROXY_PATH : `wss://${process.env.OPENSHIFT_HOST}`;
 
+  const installedServices = process.env.INSTALLED_SERVICES || '{}';
   return `window.OPENSHIFT_CONFIG = {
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -642,6 +642,7 @@ function getConfigData(req) {
   const masterUri = isOpenShift4() ? OPENSHIFT_PROXY_PATH : `https://${process.env.OPENSHIFT_HOST}`;
   const wssMasterUri = isOpenShift4() ? OPENSHIFT_PROXY_PATH : `wss://${process.env.OPENSHIFT_HOST}`;
 
+  const installedServices = process.env.INSTALLED_SERVICES || '[]';
```
JameelB: @aidenkeating yep this value is only used for OS4. Default value should be an object.

aidenkeating:
updated

server.js
tonyxrmdavidson: Confirmed LGTM

PR #527, opened 2019-11-27
Fix walkthough state caching

aidenkeating:
@ciaranRoche this was seen on master of the installation. let's get it resolved before v1.6-rhpds


ciaranRoche:
@aidenkeating verified and ran as expected

PR #534, opened 2019-12-11
Updates to expandable datalist

dlabaj:
Updated the expandable data list with datalist content.

Desktop:
Screen Shot 2019-12-12 at 3 59 22 PM

Mobile:
Screen Shot 2019-12-12 at 3 59 34 PM


package.json
```diff
@@ -5,9 +5,9 @@
   "proxy": "http://localhost:5001/",
   "dependencies": {
     "@fortawesome/fontawesome-free": "5.7.2",
-    "@patternfly/patternfly": "2.40.13",
-    "@patternfly/react-core": "3.120.8",
-    "@patternfly/react-icons": "^3.14.19",
+    "@patternfly/patternfly": "2.43.1",
```
mfrances17: I can't remember exactly, but I seem to remember a concrete reason why we did not want to go beyond a certain version of PF. I thought it was something you had found, something was not working right so we found a sweet spot of upgrading enough to fix the security vulnerability but not going far enough for what was breaking.

christiemolloy:
@dlabaj 
Just based on the screenshot, looks like the padding-top: 0 isn't being applied to the body correctly.
Also the padding on the left of the expandable body is too big

package.json
dlabaj: There was an issue but it was fixed in the latest october release. It was an a release between where the issue was showing up.  It was a console warning.

mfrances17: Ok sounds good.

tiffanynolan:
I'm not able to verify on a server, but the changes look as expected.

PR #538, opened 2019-12-12
INTLY-4390: Add missing static IDs for tabs to fix browser tests

mfrances17:
## Motivation
https://issues.redhat.com/browse/INTLY-4390

## What
Need to add static IDs for the new solution explorer tabs on the dashboard in order to unblock browser tests from failing. Also a text-only change, removed 'self-managed' from description on dashboard, since they are not available via the dashboard yet. Very low-risk changes.

## Verification Steps
Verify that when the tabs are inspected in the console dev window, they show id=servicesTab and id=solutionPatternsTab.

## Checklist:

## Progress
- [x] Finished task

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreatly-web-app",
-  "version": "2.20.6",
+  "version": "2.20.7",
   "private": true,
   "proxy": "http://localhost:5001/",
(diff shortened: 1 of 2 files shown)


tiffanynolan:
No verified on a server, but the code changes look to be as expected.

PR #552, opened 2020-02-12
INTLY-5090 fix openshift console link

pb82:
Fix the link to the openshift console from the applications list

### DIFF
--- a/src/components/installedAppsView/InstalledAppsView.js
+++ b/src/components/installedAppsView/InstalledAppsView.js
@@ -192,7 +192,10 @@ class InstalledAppsView extends React.Component {
             >
               <div className="integr8ly-state-ready">
                 <Button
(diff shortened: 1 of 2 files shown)


src/components/installedAppsView/InstalledAppsView.js
```diff
@@ -192,7 +192,7 @@ class InstalledAppsView extends React.Component {
             >
               <div className="integr8ly-state-ready">
                 <Button
-                  onClick={() => window.open(`${window.OPENSHIFT_CONFIG.masterUri}/console`, '_blank')}
+                  onClick={() => window.open(`${window.OPENSHIFT_CONFIG.masterUri}/dashboards`, '_blank')}
```
dlabaj: Do we need a check here for openshift 3?

pb82: Yes, good point. This still has to work on OS3. Will update.

PR #569, opened 2020-04-02
Cross console updates

dlabaj:
## Motivation
To support cross console navigation we need to be able to receive the list of installed apps on OpenShift for middleware.

## What
Updated server to return the list of available applications in OpenShift 4 and demo mode. Updated the client to support cross console navigation

## Checklist:

- [ X] Code has been tested locally by PR requester

## Progress

- [x] Finished task


package.json
```diff
@@ -84,7 +86,7 @@
     "watch:css": "yarn build:css && node-sass-chokidar --include-path ./src --include-path ./node_modules src/styles/index.scss -o src/styles/.css --watch --recursive",
     "build:js": "react-scripts build",
     "start": "node server.js",
-    "start:dev": "FUSE_URL=http://localhost:3006 LAUNCHER_URL=http://localhost:3006 ENMASSE_URL=http://localhost:3006 CHE_URL=http://localhost:3006 APICURIO_URL=http://localhost:3006 run-p -l watch:css start:local start:server",
+    "start:dev": "FUSE_URL=http://localhost:3006 LAUNCHER_URL=http://localhost:3006 ENMASSE_URL=http://localhost:3006 CHE_URL=http://localhost:3006 APICURITO_URL=http://localhost:3006 run-p -l watch:css start:local start:server",
```
mfrances17, line 89: why is this necessary for this specific PR? last time the name was changed it broke OS4. make sure to test on OS4 to verify it doesn't break anything. yes it should never be hit because it has the dev flag but... test. also, in openshift-template.yml it is still APICURIO_URL not APICURITO_URL, so unsure why this is being changed.

package.json
```diff
@@ -8,6 +8,8 @@
     "@patternfly/patternfly": "2.43.1",
```
mfrances17, line 8: bump the integreatly-web-app version

server.js
```diff
@@ -73,6 +107,15 @@ const WALKTHROUGH_LOCATION_DEFAULT = {
 const walkthroughs = [];
 let server;
 
+app.get('/services', (req, res) => {
+  res.setHeader('Content-Type', 'application/json');
+  if (process.env.INSTALLED_SERVICES) {
```
mfrances17, line 112: verify this doesn't break OS3/OS4 live system

server.js
```diff
@@ -300,89 +342,88 @@ function resolveWalkthroughLocations(locations) {
...
+        );
+        return resolve(locationResult);
+      } else if (isGitRepo(location)) {
+        const clonePath = path.join(TMP_DIR, tmpDirPrefix);
+
+        // Need to parse out query params for walkthroughs, e.g custom directory
+        const cloneUrl = generateCloneUrlFromLocation(location);
+        const repoName = getWalkthroughRepoNameFromLocation(location);
+        const walkthroughParams = querystring.parse(url.parse(location).query);
+
```
mfrances17, line 371: most of these seem like lint fixes that are finally being done so the changes seem worse than they actually are... makes it tough to discern only the changes for this specific feature. i'd rather they be handled separately for that exact reason but in this case it's fine.

package.json
dlabaj: This should be removed.

server.js
dlabaj: okay not sure how to undo the linting changes.

mfrances17: i'd just keep them in, they've bugged me for a while now anyway - server.js is exempt from linting blockage that's why it's always managed to slip through the cracks

server.js
dlabaj: Yep.

src/components/masthead/masthead.js
```diff
@@ -158,6 +159,35 @@ class Masthead extends React.Component {
...
   };
 
+  getCrossNavApps = () => {
+    if (this.state.appList === null) {
+      getAppsList().then(resp => {
+        const appEntries = [];
+        Object.entries(resp.data).forEach(app => {
+          switch (app[0]) {
+            case '3scale':
+              appEntries.push({ id: app[0], name: '3 Scale', rootUrl: app[1].Host.replace(/(^\w+:|^)\/\//, '') });
```
pb82, line 169: This should be written as `3scale` (no space between 3 and scale)

PR #570, opened 2020-04-03
Add 2.1 help urls

mfrances17:
## Motivation
https://issues.redhat.com/browse/INTLY-6524

## What
Change the help URLs for the getting started and release notes to the URLs provided by doc for the LA 2.1 release.

## Why
Help URLs have changed for 2.1.

## Verification Steps
1. On an OpenShift 3 or local system, select the Getting Started and Release Notes help commands and verify that they open the corresponding document. 
2. On an OpenShift 4 system, select the Getting Started and Release Notes help commands and verify that they open a 404 page (docs have not been updated on the web yet). Best you can do for now is just verify that the URLs appear in the browser address field as coded, with the 2's. 

## Checklist:

## Progress
- [x] Finished task

## Additional Notes
If needed, you can use this docker image to test:
docker.io/mfrances17/dev-tutorial-web-app:latest

### DIFF
--- a/src/components/masthead/masthead.js
+++ b/src/components/masthead/masthead.js
@@ -175,9 +175,9 @@ class Masthead extends React.Component {
       riUrl = 'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/1/html-single/release_notes/';
     } else {
       gsUrl =
-        'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/2.0/html-single/getting_started_with_red_hat_managed_integration_2.0/';
+        'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/2/html-single/getting_started_with_red_hat_managed_integration_2/';
       riUrl =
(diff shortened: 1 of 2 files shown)


src/components/masthead/masthead.js
```diff
@@ -175,9 +175,9 @@ class Masthead extends React.Component {
       riUrl = 'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/1/html-single/release_notes/';
     } else {
       gsUrl =
-        'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/2.0/html-single/getting_started_with_red_hat_managed_integration_2.0/';
+        'https://access.redhat.com/documentation/en-us/red_hat_managed_integration/2/html-single/getting_started_with_red_hat_managed_integration_2/';
```
dlabaj, line 178: Can this be set as a configuration property / environment setting in a future release so we don't have to update this every time it changes?

mfrances17: Yep, that's the plan going forward... for the 2.0 release, the URL was inconsistent with 1.0 so it didn't make sense, but as long as they stick with this version they're using for 2.1 we should be able to automate it.

PR #571, opened 2020-04-07
INTLY-6574 AMQ Online 1.4 compatibility

pb82:
## Motivation
Compatibility fix for AMQ Online 1.4

## What
AMQ Online 1.4 does no longer report a link to the address space back to the solution explorer. This PR creates the address space URL from the known host, username and password if not present in the AddressSpace CR.

## Why
To make the solution explorer work with AMQ 1.3 and 1.4

## Verification Steps
1. The solution explorer and updated WT1 are deployed here: https://console-openshift-console.apps.cluster-pbraun-34fc.pbraun-34fc.example.opentlc.com/dashboards
2. Login as test-user01 using testing-idp, select the solution explorer and make sure you can complete WT1.

## Checklist:

### DIFF
--- a/src/services/amqOnlineServices.js
+++ b/src/services/amqOnlineServices.js
@@ -23,10 +23,10 @@ const watchAMQOnline = (dispatch, username, namespace) =>
   poll(addressSpaceDef(namespace), 5000, 2000).then(pollListener => {
     pollListener.onEvent(event => {
       if (!event || !event.metadata || !event.metadata.name === cleanUsername(username)) {
-        dispatch(getPayloadFromAddressSpace(event));
+        dispatch(getPayloadFromAddressSpace(event, username, namespace));
         return;
       }
-      const payload = getPayloadFromAddressSpace(event);
+      const payload = getPayloadFromAddressSpace(event, username, namespace);
       payload.additionalAttributes = Object.assign({}, payload.additionalAttributes, {
         'enmasse-credentials-username': cleanUsername(username),
         'enmasse-credentials-password': window.atob(genEvalPassword(username))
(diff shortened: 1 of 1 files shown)


src/services/amqOnlineServices.js
```diff
@@ -139,24 +139,32 @@ const provisionAddressSpace = (username, namespace) => {
...
   resolve(payload);
 };
 
-const getPayloadFromAddressSpace = as => {
+const buildAMQOnlineConcoleUrl = (username, namespace) => {
+  if (!window.OPENSHIFT_CONFIG.provisionedServices.amqonline) {
+    return null;
+  }
+  const baseUrl = `${window.OPENSHIFT_CONFIG.provisionedServices.amqonline.Host}`;
+  return `${baseUrl}/#/address-spaces/${namespace}/${username}/standard/addresses`;
```
davidffrench, line 164: On this @pb82 . I would assume that if the address space is a brokered address space the end of this URL would be `brokered` rather than `standard`. I am not sure if the solution explorer ever creates a brokered addressspace though, so this might not be an issue

src/services/amqOnlineServices.js
```diff
@@ -139,24 +139,32 @@ const provisionAddressSpace = (username, namespace) => {
...
+  const payload = getPayloadFromAddressSpace(event, username, namespace);
   payload.additionalAttributes = Object.assign({}, payload.additionalAttributes, {
     'enmasse-credentials-username': cleanUsername(username),
     'enmasse-credentials-password': window.atob(genEvalPassword(username))
   });
   resolve(payload);
 };
 
-const getPayloadFromAddressSpace = as => {
+const buildAMQOnlineConcoleUrl = (username, namespace) => {
```
davidffrench, line 159: Should this be called `username`? Is this not the AddressSpace name?

pb82: The address space is named after the logged in user

src/services/amqOnlineServices.js
pb82: Good point. I don't think the solution explorer can create a brokered address space, but i'll check if we can get the type of the address space somehow programmatically.

davidffrench: perfect, thanks @pb82

src/services/amqOnlineServices.js
davidffrench: perfect, thanks @pb82

src/services/amqOnlineServices.js
pb82: Updated now

davidffrench:
@pb82 I am unfortunately getting an error on Task 3 of walkthrough 1 when trying to loving to OpenShift.
image

pb82:
@davidffrench fixed by adding the user manually to the rhmi-developers group (the operator was run remotely against this cluster)

PR #576, opened 2020-04-30
add functions to update and read rhmi config

pb82:
Adds a new service (rhmiConfigService) with two exported functions:

* *getCurrentRhmiConfig*: reads the current rhmi config object. This will return a promise containing an object adhering to this spec: https://github.com/integr8ly/integreatly-operator/blob/master/deploy/crds/integreatly.org_rhmiconfigs_crd.yaml

* *updateRhmiConfig*: updates the modified rhmi config object to allow the UI to manage this configuration.

### DIFF
--- a/src/services/rhmiConfigServices.js
+++ b/src/services/rhmiConfigServices.js
@@ -1,0 +1,43 @@
+import { findOpenshiftResource } from '../common/openshiftHelpers';
+import { update } from '../services/openshiftServices';
+import { rhmiConfigDef, rhmiConfigResource } from '../common/openshiftResourceDefinitions';
+
+const configName = 'rhmi-config';
+const configNamespace = 'redhat-rhmi-operator';
+
+/**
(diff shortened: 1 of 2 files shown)


src/services/rhmiConfigServices.js
```diff
@@ -0,0 +1,45 @@
...
+ * updateRhmiConfig(config).then(newConfigValues => {
+ *   // newConfigValues represents the updated config
+ * }).catch(err => handleError);
+ *
+ * @param config
+ * @returns {Promise<AxiosResponse<any>>}
+ */
+const updateRhmiConfig = config => {
+  return update(rhmiConfigDef(configNamespace), config);
+};
```
CathalOConnorRH, line 43: Looks like this is failing the test. 
try something like 
```suggestion
const updateRhmiConfig = config => update(rhmiConfigDef(configNamespace), config);
```

PR #591, opened 2020-06-03
chore(patternfly): bump for RC2

redallen:
Bump to latest RC2 versions and run pf-codemods.

This is for reference only. Merge at your own risk!


redallen:
Glad to hear the upgrade went well @mfrances17 ! Closing.

PR #594, opened 2020-06-15
Added ability to get server time.

dlabaj:
Added ability to get server time by accessing currentTime endpoint.


pb82:
@dlabaj @mfrances17 Yes, this is probably no longer required because the value of UTC will be the same regardless of the timezone (client vs. cluster).

dlabaj:
Closing PR as this feature is no longer needed.

PR #597, opened 2020-06-29
add permission check for post endpoint

pb82:
Only admins (cluster or dedicated) should be able to modify user walkthrough content. Retrieving the content is not restricted.

Verification steps:

1. Open the log tab of the solution explorer pod in one tab
2. Login to the solution explorer as customer-admin in a second tab
3. Open the settings window and submit a custom walkthrough
4. There should be a message in the logs: `Granting access to <user name>`
5. Copy the post request to `user_walkthroughs` as a fetch or curl from the browser console
6. Modify the access_token of the request and re-submit it
7. You should get a 401 response

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -9,6 +9,7 @@ module.exports = {
     'import/no-dynamic-require': 'off',
     'global-require': 'off',
     'react/jsx-curly-brace-presence': 'off',
-    'react/no-unused-prop-types': 'off'
+    'react/no-unused-prop-types': 'off',
+    'camelcase': ['error', { 'allow': ['access_token', 'auth_token'] }]
   }
 };
--- a/server.js
+++ b/server.js
@@ -4,7 +4,7 @@ const url = require('url');
 const fs = require('fs');
 const asciidoctor = require('asciidoctor.js');
 const Mustache = require('mustache');
-const { fetchOpenshiftUser } = require('./server_middleware');
+const { requireRoles } = require('./server_middleware');
 const giteaClient = require('./gitea_client');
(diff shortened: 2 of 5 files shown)


Boomatang:
I have verified the functionality works as explained in the steps above but I did find one issue that I think needs to be addressed before this can be merged. 

If you try connect to the server with an invalid token, a 500 response is given. I believe this should be a 401 response.

To cause this issue I ran 
`curl -H "X-Forwarded-Access-Token: <INVALID TOKEN>" -X POST -d '{}' https://solution-explorer.apps.<CLUSTER URL>/user_walkthroughs`

Logs in the pods showed
```
Error: Request failed with status code 401
    at createError (/opt/app-root/src/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/opt/app-root/src/node_modules/axios/lib/core/settle.js:17:12)
    at IncomingMessage.handleStreamEnd (/opt/app-root/src/node_modules/axios/lib/adapters/http.js:237:11)
    at IncomingMessage.emit (events.js:203:15)
    at endReadableNT (_stream_readable.js:1145:12)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```
and the return message in curl was 
```
<!DOCTYPE html>
<html lang="en">
<head>
<meta charset="utf-8">
<title>Error</title>
</head>
<body>
<pre>Internal Server Error</pre>
</body>
</html>
```

pb82:
@Boomatang updated the return codes now. It should now return 401 instead of 500.

Boomatang:
@pb82 I have looked over the changes and it works how I would expect it to work. In my view this can be merged.

Issue #599, opened 2020-07-09
need to do a hard refresh before you can see any SolX content


pwright:
`WALKTHROUGH_LOCATIONS=https://github.com/integr8ly/tutorial-web-app-walkthroughs yarn start:dev`

should 'just work', but it produces a blank screen

hguerrero:
This problem shows up when using latest version from master. I works if I switch to tag `2.23.0`, however you still need to make a hard reload of the page.

pwright:
I believe there are two issues, this one relates to the requirement for Ctrl-F5, and is not related to any version..
I don't have enough info to describe the @hguerrero  issue, but it went away by reverting to 2.23.0

mfrances17:
This is a day 1 issue, when running locally you would always need to refresh the browser. Not saying it doesn't need to be fixed, but it's not something that was recently introduced.

PR #601, opened 2020-07-10
Update README.adoc #599

pwright:
## Motivation
address #599

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -24,7 +24,7 @@ This web application provides the front door into the Inetgreatly initiative. It
 
 == Local Development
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -24,6 +24,8 @@ This web application provides the front door into the Inetgreatly initiative. It
 
 == Local Development
 
+IMPORTANT: The webapp will automatically open (http://localhost:3006) in your browser, however you must hard refresh the page before any content is rendered
```
mfrances17, line 27: if you want to use however, it should be used as:

...in your browser; however, you must

also missing a period at end

pwright: @mfrances17 thanks, and doh!

PR #605, opened 2020-07-27
expose upgrade data

pb82:
expose upgrade data


server.js
```diff
@@ -144,6 +144,12 @@ app.get('/about', (_, res) => {
   });
 });
 
+app.get('/upgrade_data', (_, res) => {
+  return res.json({
```
mfrances17, line 148: remove the return here

PR #612, opened 2020-08-06
APPDUX-380: Prevent duplicate backup and maintenance schedule times

mfrances17:
## Motivation
https://issues.redhat.com/browse/APPDUX-380

## What
The backup and maintenance window times should not be the same time as they would conflict with each other. In the current UI, this is prevented if you try to select the time that is already set. However, if you change both the backup time and the maintenance window time, the UI will allow you to save new duplicate times. This fix checks for that and if they both new values are the same, prevents the save, disables the Save button, and pops up a notification message.

## Why
To be more consistent with the design, and to prevent users from saving duplicate backup and maintenance window times.

## Verification Steps
1. Go to Settings > Managed Integration schedule tab.
2. Select a time in the Start time for your backups dropdown.
3. Select the same time in the Day and start time for your maintenance dropdown.
4. Verify that you stay on the same page (don't navigate back to Home), the Save button is disabled, and a notification displays and explains that you cannot set the same time for both.

## Checklist:

## Progress
- [x] Finished task

## Additional Notes
You may use my cluster to verify:
https://solution-explorer.apps.cluster-uxddev-9f0b.uxddev-9f0b.example.opentlc.com/

Or add the following docker image of the solution explorer to your own cluster:
docker.io/mfrances17/tutorial-web-app:appdux-380

Screen cap:
conflict

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreatly-web-app",
-  "version": "2.27.2",
+  "version": "2.27.3",
   "private": true,
   "proxy": "http://localhost:5001/",
   "dependencies": {
--- a/src/pages/settings/settings.js
+++ b/src/pages/settings/settings.js
@@ -310,33 +310,39 @@ class SettingsPage extends React.Component {
     buTime = this.convertTimeTo24Hr(buTime);
     maintTime = this.convertTimeTo24Hr(maintTime);
 
-    this.setState({ canSave: false });
+    if (buTime === maintTime) {
+      window.localStorage.setItem('showSettingsConflictAlert', 'true');
+      this.setState({ showSettingsConflictAlert: true });
+      this.setState({ canSave: false });
+    } else {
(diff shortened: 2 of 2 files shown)


phcox:
This is definitely working as designed, but I've noticed an issue - because the user has to scroll to the bottom of the page to save the updated schedule, it's possible that he/she will simply hit the adjacent Cancel button without seeing the error alert at the top of the page, and thereby exit the Settings thinking that the times have been scheduled as desired. This might be a small probability scenario, but could have a non-trivial impact. Can you force a scroll up to the top of the page on Save if there's a scheduling overlap?

mfrances17:
> Can you force a scroll up to the top of the page on Save if there's a scheduling overlap?
@phcox i've added a scroll up to the alert, and i've tested it in chrome and firefox, seems to do the trick. please verify on my test cluster that it meets your design expectations.

PR #613, opened 2020-08-11
Moved raw user data to variable instead of local storage.

dlabaj:
Removed storage of raw user data from local storage, instead we are now storing it in a variable.

Verified that user raw data is no longer stored in local storage and that application still works correctly.

Screen Shot 2020-08-12 at 12 55 39 PM

### DIFF
--- a/src/services/openshiftServices.js
+++ b/src/services/openshiftServices.js
@@ -3,6 +3,7 @@ import ClientOAuth2 from 'client-oauth2';
 import { isOpenShift4, getMasterUri, getWSMasterUri } from '../common/openshiftHelpers';
 
 const KIND_ROUTE = 'Route';
+let openShiftUser;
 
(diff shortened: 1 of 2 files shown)


src/services/openshiftServices.js
```diff
@@ -143,11 +144,11 @@ const getUser = () => {
  */
 const setUser = user => {
   if (!user) {
-    window.localStorage.setItem('OpenShiftUser', null);
+    openShiftUser = null;
     window.localStorage.removeItem('currentUserName');
     return;
   }
-  window.localStorage.setItem('OpenShiftUser', JSON.stringify(user));
+  openShiftUser = JSON.stringify(user);
```
mfrances17, line 151: Code looks good

dlabaj:
@jenny i fixed the progress issue.  I'm not seeing the log out issue either.

PR #614, opened 2020-09-08
remove upgrade and maintenance window settings

pb82:
## Motivation
https://issues.redhat.com/browse/APPDUX-423

## What
Remove the settings page for upgrade and maintenance window options

## Why
Restrict options so no upgrade / maintenance window outside of SRE coverage can be selected

## How
Remove the settings tab and homepage alert

## Verification Steps
1. Run the webapp locally: `yarn run start:dev`
2. Open the settings window: you should only see the tab to configure the solution patterns

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "integreatly-web-app",
-  "version": "2.27.3",
+  "version": "2.28.0",
   "private": true,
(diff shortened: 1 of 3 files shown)


CathalOConnorRH:
LGTM

